PR #73, opened 2018-04-12
Scaling group and scaling drivers

achimnol:
Let's implement policy-based scalers and scaling groups.

Concept:
* On clouds
  - Agents are assigned instance metadata describing their scaling group.
  - The scaling driver of each group runs its own auto-scaling policy and algorithm. It may be a proxy of vendor-specific auto-scaling APIs.
  - The scaling driver may be a proxy of other container orchestration framework such as Kubernetes.
* On premise
  - The default policy is "fixed".
  - Let's the users be able to provide their own plugins for external batch job schedulers such as:
    - https://github.com/IBMSpectrumComputing/lsf-python-api
    - http://www.adaptivecomputing.com/products/torque/
    - http://www.arc.ox.ac.uk/content/torque-job-scheduler
    - etc.

Policy options:
* image auto-update
  - on/off
  - update check interval
* agent auto-update
  - on/off
  - idle wait time after last kernel termination on agent
* container termination policy
  - auto-terminate after specified idle time / persistent
* external agents
  - on/off
  - public key (external agents should provide the auth message encrypted using its own private key)
  - limit (max number of allowed external agents)

Batch job functions:
* Get the job list and their status
* Submit a job (input + output + program + resource spec)
  - `JobSpecAdaptor`
* Get the cluster node list
* Get/set the scheduler configurations
* (Optional) Cancel / interrupt a job

Related:
* lablup/backend.ai-agent#5 — container termination policy (idle / persistent / etc.)
* lablup/backend.ai-agent#39 — on/off option for self-autoupgrade
* https://github.com/lablup/backend.ai-agent/projects/2 — on/off option and auth config to accept external agents


Zeniuus:
I think I need some more time to design interfaces. Please wait for my PR review request through offline.

achimnol:
* I agree with you that the scaler should be executed before the scheduler for the same single event.

We also need to consider some corner cases like this:

* A session creation request may demand a resource specification *impossible to schedule* with current hardware configurations *even after scaling out.* For instance, when we use only one type of GPU instances having 1.0 GPU share for each, allocating 1.5 GPU share for a single-container session is impossible no matter how many instances the scaler adds.
* In this case, we should let the user (client) know the request is impossible to schedule without waiting for the pending-queue timeout.

Zeniuus:
Suggestion: I think `AbstractScalingDriver` contains two separate parts: scaling logic and vendor-specific implementation. For instance, `scale_out` and `get_scale_out_info` are independent of which vendor is used, and `add_agents` and `get_ami_id` are independent of scaling logic.

So, I think it is good to separate AbstractScalingDriver into AbstractScalingDriver and VenderScalingDriverMixin. Is it seemed to be over-engineering?

achimnol:
At first I thought it would be sufficient to have vendor-specific implementation and the scaling logic together. But I think if you can separate two, then it is much better to do so!

Zeniuus:
I have a doubt about total separation of `AbstractScalingDriver` and `AbstractJobScheduler`.

Consider the following case. `AbstractJobScheduler` implements a policy that assigns kernels with no GPU usage to only t2.2xlarge instances, never to p2.xlarge instances, to minimize the usage of expensive p2.xlarge instance, whereas `AbstractScalingDriver` determines to scale out when the total sum of resources of any type of agents is insufficient to deal with all pending `SessionCreationRequest`s. And currently, there is an only one p2.xlarge instance with no session.

In this case, if session creation request with no GPU usage comes in, then `AbstractScalingDriver` would not scale out because there are enough CPU resource shares in the p2.xlarge instance, and `AbstractJobScheduler` would not schedule the request since there is no t2.2xlarge instance. So every CPU-only kernels will be pending infinitely.

But adding a method like `calc_required_instances` to `AbstractJobScheduler` does not seem to be right since detail logic of how to scale out/in is the responsibility of `AbstractScalingDriver`.

Do you have any idea for this issue?

achimnol:
> I have a doubt about total separation of `AbstractScalingDriver` and `AbstractJobScheduler`.
> 
> Consider the following case. `AbstractJobScheduler` implements a policy that assigns kernels with no GPU usage to only t2.2xlarge instances, never to p2.xlarge instances, to minimize the usage of expensive p2.xlarge instance, whereas `AbstractScalingDriver` determines to scale out when the total sum of resources of any type of agents is insufficient to deal with all pending `SessionCreationRequest`s. And currently, there is an only one p2.xlarge instance with no session.
> 
> In this case, if session creation request with no GPU usage comes in, then `AbstractScalingDriver` would not scale out because there are enough CPU resource shares in the p2.xlarge instance, and `AbstractJobScheduler` would not schedule the request since there is no t2.2xlarge instance. So every CPU-only kernels will be pending infinitely.

Good catch for yet another corner case!

> But adding a method like `calc_required_instances` to `AbstractJobScheduler` does not seem to be right since detail logic of how to scale out/in is the responsibility of `AbstractScalingDriver`.
> 
> Do you have any idea for this issue?

There are many potential ways to resolve this issue:

* Strictly treat all resources equivalently, meaning that even CPU-only sessions may be scheduled on GPU instances. ( == There will be no such scheduler you exemplified!)
  - (+) Simple and consistent.
  - (-) We cannot use policies like you have exemplified.
* Annotate each request with "favored" instance type (e.g., "t2" or "p2") when first checked by the scheduler (this may be configuration of either the scheduler or images) and let the scaler to consider them when reading the pending request queue.
  - This way, virtually the list of resource types becomes something like "t2.cpu", "t2.mem", "p2.cpu", "p2.mem", and "p2.cuda" whereas each requests have a subset of those types.
  - (+) No interface changes.
  - (-) A scaler implementation may ignore such annotations since they are just annotations. This means that there is no guarantee of working for arbitrary combination of scheduler/scaler, just like Python's type annotation.
* Integrate the scheduler and scaling driver into one interface.
  - (+) We can do anything with them without unexpected breaking.
  - (-) It may become future obstacles to implement 3rd-party scheduler plugins since they also need to implement the scaling driver part.

Zeniuus:
Thank you for suggesting concrete possible solutions! As long as not wasting time, I'll try to figure out this issue more to find a better solution in aspect of both modularization and stability. I think we can find a good solution... 🤔

Zeniuus:
- To implement minimum prepared shares, I introduced a concept of 'share unit'. AbstractScalingDriver can specify its minimum prepared shares by defining various share unit and its required number. For example, to express 10 CPU + 10 GiB memory, we can define a share unit as 1 CPU + 1 GiB memory and set its required number as 10. You can define many different share unit at the same time to handle different situations, as specified in `BasicScalingDriver.min_prepared_shares`.
- When scaling out, BasicScalingDriver generates `buffer_jobs` to represent minimum prepared shares. The result of scaling out should be able to run both `pending_jobs` and `buffer_jobs`. In this way, BasicScalingDriver can always prepare for spare resources.
- I added a new ADT class, SessionCreationJob, for representing job for a scheduler. It is a just lighter version of SessionCreationRequest. It is useful when creating a 'dummy' jobs such as `buffer_jobs`.

achimnol:
This is now considered replaced with #167.
Adding auto-scaling driver will be done as a separate issue in the futureu.

Issue #53, opened 2017-11-08
API New: User API for Querying Image Metadata


achimnol:
The list of available kernels and their respective metadata should be read from the Backend.AI manager instead of hard-coded in the client-side.

 * [x] Write methods in `ai.backend.gateway.etcd.ConfigServer` to retrieve image metadata.
 * [x] Add a new GraphQL schema type `images`

PR #94, opened 2018-10-08
Feature/image metadata api

Zeniuus:
Pull request for #53.

Currently `ai.backend.gateway.etcd.ConfigServer::get_image_data` returns only image hash, cpu, gpu, memory slots.

Also,
- querying metadata of multiple, specified images (e.g. `name=['image1:tag1, 'image2:tag2', ... ]`)
- `private` optional parameter

are not implemented yet.

Please give me comments about detail spec of this API.

### DIFF
--- a/src/ai/backend/gateway/admin.py
+++ b/src/ai/backend/gateway/admin.py
@@ -135,7 +135,9 @@ class QueryForAdmin(graphene.ObjectType):
         sess_id=graphene.String(required=True),
         status=graphene.String())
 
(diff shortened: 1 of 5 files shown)


achimnol:
Could you update this PR to match with #108?

achimnol:
Replaced with #127 as we decided to make the docker registries as the authorative source of all kernel metadata, while keeping the etcd data as cache. #127 also restructures the etcd values.
So it is no longer possible to apply this PR. 😭 Closing...

Issue #91, opened 2018-10-05
Generalize hard-coded folder host in vfolder API


achimnol:
* [x] Let [the parameters of the vfolder create API](https://github.com/lablup/backend.ai-manager/blob/9dd2839/src/ai/backend/gateway/vfolder.py#L134) to have an explicit "folder host" argument.
   - [x] When not specified, use a default folder host configured in etcd. If no default is configured, raise an exception `VFolderCreationFailed`. Derive the exception class from `aiohttp.web.HTTPServiceUnavailable` and `ai.backend.gateway.exceptions.BackendError` so that it conforms with RFC7807-style error reports.
   - [x] Validate the given folder host argument using the first-depth directory list of the vfolder root (e.g., `/mnt/vfroot/`)
 * [x] Update [the client](https://github.com/lablup/backend.ai-client-py) to provide:
   - [x] an argument specifying the folder host in `BaseVFolder._create()` method
   - [x] a CLI option to specify the folder host in `vfolder create` command
 * [x] Update [the automatic dev-setup script](https://github.com/lablup/backend.ai/blob/028d17/scripts/install-dev.sh#L369) to use an example name (e.g., `local`) instead of `azure-shard01` and let it configure the default vfolder host as it (`local`).
 * [x] Add an API to retrieve the list of available vfolder hosts against a given access key.
 * [x] Update [documentation](https://github.com/lablup/backend.ai/tree/master/docs) about virtual folder APIs

PR #96, opened 2018-10-08
General vfolder host

Zeniuus:
Pull request for #91.

* Add a new logic to decide folder_host
* Change the logic to check duplicate vfolders

### DIFF
--- a/src/ai/backend/gateway/exceptions.py
+++ b/src/ai/backend/gateway/exceptions.py
@@ -104,6 +104,11 @@ class KernelAlreadyExists(web.HTTPBadRequest, BackendError):
                   'a different runtime type (language).'
 
 
+class VFolderCreationFailed(web.HTTPBadRequest, BackendError):
+    error_type  = 'https://api.backend.ai/probs/vfolder-creation-failed'
+    error_title = 'Virtual folder creation has failed.'
+
+
 class FolderNotFound(web.HTTPNotFound, BackendError):
     error_type  = 'https://api.backend.ai/probs/folder-not-found'
     error_title = 'No such virtual folder.'
@@ -222,8 +227,3 @@ class KernelRestartFailed(web.HTTPInternalServerError, BackendAgentError):
 class KernelExecutionFailed(web.HTTPInternalServerError, BackendAgentError):
     error_type  = 'https://api.backend.ai/probs/kernel-execution-failed'
     error_title = 'Executing user code in the kernel has failed.'
-
-
-class VFolderCreationFailed(web.HTTPServiceUnavailable, BackendError):
-    error_type  = 'https://api.backend.ai/probs/vfolder-creation-failed'
-    error_title = 'VFolder creation has failed.'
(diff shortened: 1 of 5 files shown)


src/ai/backend/gateway/vfolder.py
```diff
@@ -134,6 +136,16 @@ def vfolder_check_exists(handler):
     params = await request.json()
     log.info(f"VFOLDER.CREATE (u:{access_key})")
     assert _rx_slug.search(params['name']) is not None
+    # Resolve host for the new virtual folder.
+    folder_host = params.get('host', None)
+    if not folder_host:
+        folder_host = \
+            await request.app['config_server'].etcd.get('volumes/_default_host')
+        if not folder_host:
+            raise VFolderCreationFailed
```
achimnol: It would be better to add a short message to the exception describing the situation, e.g., "You must specify the vfolder host because the default vfolder host is not configured."

src/ai/backend/gateway/vfolder.py
```diff
@@ -134,6 +136,16 @@ def vfolder_check_exists(handler):
...
+    if not (request.app['VFOLDER_MOUNT'] / folder_host).is_dir():
+        raise VFolderCreationFailed
```
achimnol: Same here, let's add a message like "Invliad vfolder host."

tests/conftest.py
```diff
@@ -77,6 +84,9 @@ def prepare_and_cleanup_databases(request, test_ns, test_db, folder_mount):
     args = Namespace(key='volumes/_mount', value=str(folder_mount),
                      etcd_addr=etcd_addr, namespace=test_ns)
     put(args)
+    args = Namespace(key='volumes/_default_host', value=str(folder_host),
+                     etcd_addr=etcd_addr, namespace=test_ns)
+    put(args)
```
achimnol: 👍

src/ai/backend/gateway/exceptions.py
```diff
@@ -222,3 +222,8 @@ class KernelRestartFailed(web.HTTPInternalServerError, BackendAgentError):
 class KernelExecutionFailed(web.HTTPInternalServerError, BackendAgentError):
     error_type  = 'https://api.backend.ai/probs/kernel-execution-failed'
     error_title = 'Executing user code in the kernel has failed.'
+
+
+class VFolderCreationFailed(web.HTTPServiceUnavailable, BackendError):
```
achimnol: At first I thought this error suits for `HTTPServiceUnavailable`, but I'm more inclined to use `web.HTTPBadRequest`, because this error is often caused by misconfiguration of the server (e.g., missing configuration of the default vfolder host) or the invalid parameters (e.g., passing a non-exisitng vfolder host).
Let's change it to base on `web.HTTPBadRequest`.

Also we need to move the declaration of this exception above `AgentError` which represents agent-side errors, since `VFolderCreationFailed` is a type of manager-side errors.

Zeniuus:
I wonder if I should change naming of other vfolder-related `Exceptions` (`FolderNotFound`, `FolderAlreadyExists`)
They currently starts with `Folder`, not `VFolder`, and so as `error_type`.

achimnol:
I think it is better to change as you noted: `FolderXXX` exceptions -> `VFolderXXX` exceptions.
Since currently they are only used inside the manager and the client only sees serialized error data, there would be no impacts to other Backend.AI components.

PR #97, opened 2018-10-12
Load app plugins through entrypoint

Zeniuus:
Detect app plugins through entrypoint, not by scanning modules in `BACKEND_EXTENSIONS`.

Users can disable installed app plugins through `--disable-plugins` cli option.

### DIFF
--- a/src/ai/backend/gateway/server.py
+++ b/src/ai/backend/gateway/server.py
@@ -334,7 +334,7 @@ async def server_main(loop, pidx, _args):
         subapp_mod = importlib.import_module(pkgname, 'ai.backend.gateway')
         init_subapp(getattr(subapp_mod, 'create_app'))
 
-    app_plugin_entry_prefix = 'backendai_app_v10'
+    app_plugin_entry_prefix = 'backendai_webapp_v10'
     if app['config'].disable_plugins:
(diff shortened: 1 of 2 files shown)


src/ai/backend/gateway/server.py
```diff
@@ -331,6 +328,23 @@ def _get_legacy_handler(handler, app, api_version):
                 handler = _get_legacy_handler(r.handler, subapp, version)
                 app.router.add_route(r.method, legacy_path, handler)
 
+    for pkgname in subapp_pkgs:
+        if pidx == 0:
+            log.info('Loading module: %s', pkgname[1:])
+        subapp_mod = importlib.import_module(pkgname, 'ai.backend.gateway')
+        init_subapp(getattr(subapp_mod, 'create_app'))
+
+    app_plugin_entry_prefix = 'backendai_app_v10'
```
achimnol: Though the code uses simply "app" to represent `aiohttp.web.Application`, I think it's better to use `backendai_webapp_v10` here to distinguish this with other types of apps that might be added in the future.
Just a little more clarification. :wink:

Zeniuus: Thank you for the feedback!

By the way, is there any reason for using "version 10"?

PR #108, opened 2018-10-29
Support dual docker registry

Zeniuus:
Pull request for [lablup/backend.ai-agent#82](https://github.com/lablup/backend.ai-agent/issues/82).

How to select registry: (in order of higher priority)
1. Registry specified by 'lang' params in POST /kernel/create request
2. Registry specified when starting manager
3. Default registry, 'lablup'


Zeniuus:
I've updated alembic revision file! I missed that resizing length limit of String col is not automatically detected.

Also, I fixed a bug that registry name with two or more `/` does not work.

@achimnol

Issue #106, opened 2018-10-27
Freeze mode for maintenance


achimnol:
Let's add a freeze mode which can be activated by a manager CLI command for maintenance.
In the freeze mode, the following operations are prohibited:
* Creation of new kernels
* Creation/modification of users & keypairs
* Creation/modification of virtual folders
* (Future) Creation/modification of user policies & scaling-groups

Existing kernel sessions are still executed and users can interact with them.

If `--wait` option is given to the command, let the command to wait until there is no running sessions and return.

* [x] Add a new API exception `ServerFrozen` which inherits `HTTPServiceUnavailable`
* [x] Add a new API endpoint to check if the server is frozen, which returns a simple JSON like `{"status": "frozen" | "running", "active_sessions": 0}`.
  - Checking the frozen status from the etcd should be done once every 30 seconds.
    - [x] Extend the config server to provide a frozen checker method, which uses `aiotools.lru_cache` with 30-sec expiration.
    - [ ] TODO: Could we make a distributed agreement procedure to ensure that all threads/instances of the gateway are actually frozen? (maybe counting the number of frozen threads using atomic updates of etcd values)
* [x] Add a new manager CLI command to `freeze` and `unfreeze`, which stores a boolean status in the etcd.
  - [x] Let the command to wait until the frozen-status API endpoint to return "frozen" status by periodic polling (e.g., on every 3 seconds).
  - [x] If `--wait` option is given, further wait until the value of `"active_sessions"` becomes zero.
* [ ] Change the admin mutation APIs and REST APIs that incurs the operations mentioned above to return `ServerFrozen` error if the server state is frozen, using the config server method written above.

achimnol:
Let's also add another boolean option to *force kill* all running sessions: `--force-kill`.

PR #109, opened 2018-10-30
Freeze mode for manager

Zeniuus:
Pull request for #106.

I ensure all managers share same frozen status by `AsyncEtcd::watch`ing same key in all manager processes. When update of 'manager/status' key in etcd server is detected, each process expires its `ConfigServer::get_manager_status` cache.


src/ai/backend/gateway/manager.py
```diff
@@ -0,0 +1,128 @@
...
+log = BraceStyleAdapter(logging.getLogger('ai.backend.gateway.manager'))
+
+
+class ManagerStatus(enum.Enum):
+    RUNNING = 'running'
+    FROZEN = 'frozen'
+
+    @classmethod
+    def is_member(cls, value):
+        return any(value == item.value for item in cls)
```
achimnol: I think this is an unncessary boilerplate.
We can check this membership by checking if `ManagerStatus(some_string)` raises `ValueError` or not.

src/ai/backend/gateway/manager.py
```diff
@@ -0,0 +1,128 @@
...
+    def wrapper(handler, gql=False):
+
+        @functools.wraps(handler)
+        async def _wrapped(request):
+            if gql:
+                # Check if server is frozen only for mutation queries.
+                try:
+                    params = await request.json()
+                    query = params['query']
+                    if not query.startswith('mutation'):
```
achimnol: This looks dangerous because a single GraphQL query may contain multiple queries and mutations in an arbitrary order.
Could we use graphene or its dependencies to check if the query actually contains any mutation?

src/ai/backend/gateway/exceptions.py
```diff
@@ -134,6 +134,11 @@ class InstanceNotAvailable(web.HTTPServiceUnavailable, BackendError):
     error_title = 'There is no available instance.'
 
 
+class ServerFrozen(web.HTTPServiceUnavailable, BackendError):
+    error_type  = 'https://api.backend.ai/probs/server-frozen'
+    error_title = 'The server is in frozen status.'
```
achimnol:
```suggestion
    error_title = 'The server is in frozen due to maintenance. Please try again later.'
```

tests/gateway/test_manager.py
```diff
@@ -0,0 +1,169 @@
...
+from ai.backend.manager.cli.etcd import put
+
+
+def test_manager_status_is_member():
+    assert ManagerStatus.is_member('running')
+    assert ManagerStatus.is_member('frozen')
+    assert not ManagerStatus.is_member('freeze')
+    assert not ManagerStatus.is_member('stop')
+    assert not ManagerStatus.is_member('terminated')
+    assert not ManagerStatus.is_member('')
```
achimnol: As commented in src/ai/backend/gateway/manager.py, this is unnecessary.

src/ai/backend/gateway/manager.py
Zeniuus: OK, I'll try to find related APIs!

Zeniuus:
Updated codes as you reviewed, and added `--force-kill` option.

@achimnol

achimnol:
Ooops, I missed that `admin_required` is missing in `update_manager_status()` API handler in the new `src/ai/backend/gateway/manager.py`.
Only admins can freeze the server!

Issue #107, opened 2018-10-28
Monitoring Plugins


achimnol:
Datadog / Sentry could be extracted as statistics and error monitoring plugins with simple callback APIs. Let's do this and make them optionally installable.

PR #112, opened 2018-10-31
Refactor out monitoring tools as plugins

Zeniuus:
Pull request for #107.

Abstraction of monitoring tools:
- datadog -> [`AbstractStatsMonitor`](https://github.com/lablup/backend.ai-stats-monitor/blob/master/src/ai/backend/monitor/stats/base.py)
- sentry -> [`AbstractErrorMonitor`](https://github.com/lablup/backend.ai-error-monitor/blob/master/src/ai/backend/monitor/error/base.py)

Interface to use plugin:
- [`add_plugin_args`](https://github.com/lablup/backend.ai-stats-monitor/blob/master/src/ai/backend/monitor/stats/datadog.py#L33-L39): Interface to add args needed for plugin to app's parser
- [`get_plugin`](https://github.com/lablup/backend.ai-stats-monitor/blob/master/src/ai/backend/monitor/stats/datadog.py#L42-L45): Interface to get initialized plugin

Related PRs:
- [lablup/backend.ai-agent#84](https://github.com/lablup/backend.ai-agent/pull/84)
- [lablup/backend.ai-common#3](https://github.com/lablup/backend.ai-common/pull/3)

\* Note that manager's and agent's branch depends on common's branch, so we need to rebuild after common's PR is merged.

See also:
- [lablup/backend.ai-stats-monitor repo](https://github.com/lablup/backend.ai-stats-monitor)
- [lablup/backend.ai-error-monitor repo](https://github.com/lablup/backend.ai-error-monitor)


achimnol:
The code looks good, but I have some design questions:

1. Is it possible to add multiple or zero stats/error monitors? (e.g., using Datadog and Newrelic stats monitoring plugins at the same time without changing calls to the stats monitors in manager/agents)
2. How about refactoring out "plugin discovery and loading" mechanisms as one of the common module?

Zeniuus:
1. Now I understand what you meant by 'registry'. :D
2. I totally agree with your idea! I was wondering if implementing plugin-discovery function in common module is over-engineering or not.

Zeniuus:
I've implemented some helper classes & functions for installing plugins in [common repo](https://github.com/lablup/backend.ai-common/pull/3/commits/badec12c90d29a417ac3c7fdb1fe0a502ce13c16).

- `PluginRegistry`: Simple registry to propagate method call to multiple plugins
- `install_plugins`: Automatically install `plugins` to `app`
- `add_plugin_args`: Automatically add `plugins`' args to app's `parser`
- `discover_plugins`: Automatically detects `plugins`' entry points. Note that it does not install plugins, but only detects them and return entry points. This is for `backendai_webapp_v10` plugins, whose installation logic is much more complicated so that it cannot be automated.

@achimnol

Issue #105, opened 2018-10-26
Auto-detection and auto-initialization of Docker Swarm cluster


achimnol:
To implement private overlay networks for compute session with multiple bundled containers (#32), the manager also needs a local docker daemon and this daemon must be the master of a swarm cluster.

* [ ] Detect if the docker swarm cluster is initialized already or not.
  - [ ] If not, automatically initialize it.
    - [ ] To prevent unintended service interruptions, the manager node should be exempted from running containers in the swarm (i.e., "manager-only node"), though Backend.AI manages the container lifecycles.
* [ ] (For each scaling group:) Store the swarm master address and the swarm token to the etcd.
* [ ] Let the agent make their docker daemons to detect & join the swarm cluster when starting up using the config stored in the etcd.

Reference:
* https://docs.docker.com/engine/swarm/admin_guide/

Relevance to the scaling group:
* Each scaling group corresponds to a single Swarm cluster (if it is a standard Backend.AI scaling groups, instead of proxy-plugin to 3rd-party batch job schedulers).
* Since only one Docker daemon can run on a single Backend.AI manager node, the default scaling group may have the Swarm manager running along with the Backend.AI manager on the same node, but other additional scaling groups should have their own Swarm manager nodes separately.

achimnol:
We decided to use classic-mode swarm for stability.
Replaced with lablup/backend.ai#78.

Issue #32, opened 2017-04-07
Bundled containers for a kernel session


achimnol:
For large-scale computations, sometimes we need to run multiple containers on different hosts for resource aggregation and distributed/parallel processing.

In the past, this was very difficult to implement because Docker's networking was limited to mounting a container to another via an hostname alias (`--link`). Now, it's 2017, and Docker offers a nice distributed coordination called "Swarm".

Docker Swarm uses the Raft algorithm to share node information and any new Docker daemon can join to an existing Swarm via host:port and a secret token. Once joined, their containers can be connected to volatile overlay networks created and destroyed at runtime.

Let's try this and support multi-container distributed computing!

PR #115, opened 2018-11-18
Multi-container sessions

achimnol:
PR for #32. Assuming that #105 is done (either manually configured or implemented).

 * A multi-container session consists of a `master` kernel and many `worker` kernels.
 * The kernels reside in the same overlay network managed by Docker Swarm.

**NOTE:** You must have the Docker daemon installed in the manager node as well (which was not a requirement previously!) and all agents should join to the swarm cluster hosted by the manager.

TODOs:
* [x] Create/delete multi-container sessions with role = "master" / "worker"
  - Since `sess_id` can be reused, I introduced a new `bundle_id` field that plays as the unique key for the session. So we have three different identifiers:
    - `sess_id` – represents client-side view of a session (`get_or_create` API reuses existing session if the same session ID is given again while it's running.)
    - `bundle_id` – corresponds to a specific launch of session
    - `kernel_id` – corresponds to a container (`container_id`)
* [x] Create/delete overlay networks attached to the sesison
  - [ ] Need to fix "network is in used by task" error intermittently happening when session clean up
* [ ] Differentiate master/worker images for a single kernel (e.g., Hadoop namenode/datanode)
* [ ] Make a way to specify master/worker commands separately for the execute API


achimnol:
Observe problem:

* After observing the following error when destroying a network, next session launches with multiple containers does not work -- the container creation hangs and time-out is reported by the docker. 
```
Traceback (most recent call last):
  File "/Users/joongi/Projects/Lablup/backend.ai/manager/src/ai/backend/gateway/kernel.py", line 175, in kernel_terminated
    await app['registry'].mark_kernel_terminated(kernel_id)
  File "/Users/joongi/Projects/Lablup/backend.ai/manager/src/ai/backend/manager/registry.py", line 1024, in mark_kernel_terminated
    await network.delete()
  File "/Users/joongi/.pyenv/versions/3.6.7/envs/venv-bai-manager/lib/python3.6/site-packages/aiodocker/networks.py", line 36, in delete
    method="DELETE",
  File "/Users/joongi/.pyenv/versions/3.6.7/envs/venv-bai-manager/lib/python3.6/site-packages/aiodocker/docker.py", line 166, in _query
    json.loads(what.decode('utf8')))
aiodocker.exceptions.DockerError: DockerError(400, 'rpc error: code = FailedPrecondition desc = network 15visbccz249mg1yrawcz3b76 is in use by task pmoztbno276poxphdkw0zedcp')
```

achimnol:
Maybe netwok removal errors are related to this issue: https://github.com/moby/moby/issues/31068

achimnol:
We also need to find workarounds for errors happening when creating two or more overlay networks: https://github.com/moby/moby/issues/35888

achimnol:
We decided to implement a "single-node multi-container" scenario first, with an architecture extensible to "multi-node multi-conatiner". Some of codes in this PR will be merged to #217.

Issue #111, opened 2018-10-30
Consistent resource share value scales


achimnol:
Currently API and the database uses MiB for storing memory sizes while the resource allocators use relative "share" decimal values whose unit is GiB.

The commit b8c18c6 temporarily mitigates this mismatch when handling agent/kernel memory sizes.

Let's make it consistent resource share values in v18.12 by converting all resource field values into `NUMERIC(12, 3)` standard SQL column type (the number of significant digits are 12 where the value is scaled down 10^3) which should be sufficient for storing relative resource shares.

achimnol:
Done with #127.
Resource slots can be dynamically defined by adding accelerator plugins.
There are two types of resource slots: count, bytes.
Count can be fractional (using Decimal) and all their values are stored as strings in JSON not to loose accuracy and to avoid floating point rounding errors against many times of addition and subtraction.

PR #127, opened 2019-01-24
Rewrite kernel metadata management

achimnol:
This PR is work in accordance with feature/auto-kernel-metadata in the agent/common projects.

To work with this PR on existing development setups, run:
```shell
python -m ai.backend.manager.cli etcd delete --prefix images
python -m ai.backend.manager.cli etcd put \
    config/docker/registry/lablup "https://registry-1.docker.io/"
python -m ai.backend.manager.cli etcd rescan-images -r lablup
```

To go back for previous versions, run:
```shell
python -m ai.backend.manager.cli etcd delete --prefix images
python -m ai.backend.manager.cli etcd update-images -f sample-configs/image-metadata.yml
python -m ai.backend.manager.cli etcd update-aliases -f sample-configs/image-aliases.yml
```

* [x] Extend `images` GraphQL query so that it provides min/max resource requirements, human-friendly names, and other additional information about the kernel image.
  - Now the result includes the image size in bytes and aliases targeting it as well.
* [x] Implement `alias_image`, `dealias_image`, `rescan_images` GraphQL mutations.
  Theay are also provided as manager CLI subcommands.
* [ ] Automatically generate image aliases and merge duplicated aliases, while admins can define additional aliases manually (which should have higher priority than auto-generated ones) -> postponed
* [x] Restructure "images" subpaths in etcd.
* [x] Extract and store accelerator-specific resource slot metadata from heartbeats.
* [x] Rewrite `ResourceSlot` and `ResourceSpec` classes and change the `kernels` and `agents` table schema with alembic migrations to use dynamic accelerator resource slots. (We will use JSON column types!)
* [x] Implement `etcd set-image-resource-limit` command as in the manual.
* [x] When creating kernels, determine if the agent needs to pull a new/updated image. Instruct the agent to do so if required. -> implemented using digest comparison!
* [x] Test if it can spawn compute sessions using the new kernel metadata.

Related: #111, lablup/backend.ai-agent#98


achimnol:
lablup/backend.ai-common#5 must be merged to continue.

PR #132, opened 2019-02-04
graphQL pagination for sessions, agents

tink-expo:
handles client's paginated graphQL requests to retrieve sessions and agents.
- added count_compute_sessions, count_agents.
- now compute_sessions, agents respond with requested offset and limit.


achimnol:
For backward compatibility, let's add:
```python
agent_list = AgentList(...)
```
instead of replacing the existing `agents` field in `QueryForAdmin` and `QueryForUser`.

achimnol:
For the issue about requirement of at least one field in the `Item` interface, let's use `graphene.ID` field which accepts arbitrary strings/integers as its value.

src/ai/backend/manager/models/agent.py
```diff
@@ -104,6 +108,39 @@ def from_row(cls, context, row):
...
+                       .as_scalar())
+            if status is not None:
+                status = AgentStatus[status]
+                query = query.where(agents.c.status == status)
+            result = await conn.execute(query)
+            count = await result.fetchone()
+            return count[0]
+
+    @staticmethod
+    async def load_with_limit(context, limit, offset, status=None):
```
achimnol: Let's rename this (and the other in `ComputeSession`) to `load_sliced`. :wink:

Issue #136, opened 2019-02-08
Inconsistent slot resource format


inureyes:
## Description

Value formats in slot resources (occupied_slots, available_slots) is inconsistent.

### Case 1
`occupied_slots: "{"cpu": "8", "mem": "64 GiB", "cuda.device": "1"}"`

### Case 2
`occupied_slots: "{"cpu": "1", "mem": "1g", "cuda.device": "0"}"`

### Case 3
`occupied_slots: "{"cpu": "1", "mem": "256 MiB", "cuda.device": "0"}"`

### Case 4
`occupied_slots: "{"cpu": "1", "mem": "256m", "cuda.device": "0"}"`

In JS library and GUI console, it is handled with handwritten fallback routine. (see https://github.com/lablup/backend.ai-console/commit/be5983903d33e712d11b6caaaa7857e826bb9f06 ) However, we need a polished solution / resource description rules to avoid duplicated dirty jobs.

achimnol:
I can alway numericalize all numbers, but still it is the client's job to determine which field represents bytes or count. Also to handle big integers, all values are stringified.
Are you okay with this?

achimnol:
This is now rewritten and resolved.

PR #142, opened 2019-03-27
Resource presets

achimnol:
refs: #136, #134

It is difficult to expose the exact amount of allocatable resource slots in the GUI, because there are multiple resource slots and multiple agents where it is non-trivial to define the priority of each resource slot.

As a workaround, let's use a set of "resource presets" (similar to instance types in AWS EC2) where users can easily choose one of them for new sessions in the GUI. Optionally users should be able to switch to the current slider-based resource selector as an "advanced" UI where it is not guaranteed to be successful for arbitrary combinations of resource selection.

To support this in the GUI, the manager should keep the list of resource presets so that administrators can customize them.

* Add `ResourcePreset` GraphQL object and `resource_presets` DB table.
  - Anyone can read resource presets.
  - Only admins can create/modify/delete resource presets.
* Add new REST APIs
  - `/resource/presets`: Get all resource presets
  - `/resource/check-presets`: Return the allocatability of each resource preset with all resource presets given the current access key and the current scaling group
* Also improve handling of "infinity" in resource slots


achimnol:
Currently work in progress. I will notify you (@inureyes) once completed.

PR #143, opened 2019-04-03
Split DB transactions for kernel creation

achimnol:
This has been a major cause of TimeoutError due to connection pool exhaustion during load tests.


achimnol:
Some experiments shows that:
* `docker start` (invoked via API) often has a very long delay under high loads, up to more than 60 seconds.
* Splitting DB transactions around the API invocation resolves the timeout errors completely.
* We need to improve synchronization for `keypairs.concurrency_used` -- it sometimes experiences race conditions.

achimnol:
We still need to inspect synchronization for `keypairs.concurrency_used`, but that is orthogonal to transaction splits. As the PR is proven to resolve aiopg timeout errors under high loads, we first merge it and then go into the sync problem.

PR #150, opened 2019-05-19
Feature/#148 add user model

adrysn:
I open pull request for eaiser review.

* `User` model added.
  * `password` is hashed (with bcrypt) before save.
  * `role` field can be `admin`, `user`, or `monitor`, currently.
  * Basic CRUD graphql operations.
  * Migration.
    * Create users based on `keypair.user_id` (== email) and associate `keypairs.user` to user's `uuid` field.
    * User's role is set to `admin` if one of keypairs's `is_admin` is `True`. Otherwise, user's role is `user`. No `monitor` roles for previous keypairs.
  * User's information is appended from auth middleware.
* `Domain` model added.
  * An admin not belong to any domain is regarded as super-admin currently. An admin with domain is should be recognized as a domain admin.
  * `name` should be a slug field.
  * Basic CRUD graphql operations.
    * Domain admin do not have CRUD permission for other domains. 
  * Migration.
    * `default` domain is created. All previous users are associated with the default domain.
* `Group` model added (`== Project`).
  * `name` should be a slug field.
  * `association_groups_users` maintains M2M relationship between `groups` and `users`.
  * Basic CRUD graphql operations.
    * Domain admin do not have permission to groups in other domain.
* Test cases for CRUD operations for three added models.


achimnol:
Good to see the followings:
* Use of bcrypt for user password hashing
* Fully covered alembic migrations
* Test cases

I think it would be better to merge at the current stage and split follow-up tasks again.
👍

PR #160, opened 2019-07-10
Update REST API of kernel.get_info() to be up-to-date with models/kernel.py

temirrr:
I updated the file, while making the following assumptions:
- Detailed text input is better for end-user rather than raw output message. Consequently, I formatted data like `resp['age'] = "{} milliseconds".format(age_ms)` or `resp['terminationTime'] = str(kern.terminated_at) if kern.terminated_at else "Kernel is not terminated yet"` to be more readable for endusers.
- I assumed that we won't need to return the deprecated fields anymore. E.g. "idleTimeout", etc.
- I camelcased all the keys in resp dictionary. I named all new fields myself, since I didn't find them being used somewhere yet.

Please, review it when you have time. I'll correct the mistakes as requested.

Sincerely,

Temirlan


src/ai/backend/gateway/kernel.py
```diff
@@ -353,23 +353,35 @@
...
-        # TODO: factor out policy/image info as a common repository
-        resp['queryTimeout']  = -1  # deprecated
-        resp['idleTimeout']   = -1  # deprecated
-        resp['memoryLimit']   = kern.mem_max_bytes >> 10  # KiB
-        resp['maxCpuCredit']  = -1  # deprecated
-        # Stats collected from agent heartbeats
+        age_ms = int(age.total_seconds() * 1000)
+        resp['age'] = "{} milliseconds".format(age_ms)
+        resp['creationTime'] = str(kern.created_at)
+        resp['terminationTime'] = str(kern.terminated_at) if kern.terminated_at else "Kernel is not terminated yet"
```
achimnol: As being an API, which is for *programmatic* access, this should be deliberately set to `None` (or something else that *machines* can handle and distinguish with normal values easily) if not specified. Showing user-facing messages is the job of the clients.

temirrr: @achimnol I corrected it as you requested.

Issue #119, opened 2018-11-30
API New: User-created images


achimnol:
Let's users be able to stop & commit their sessions as a new Docker image, and reuse them afterwards.

At first, we will support this only in single-container sessions.

achimnol:
In 19.03.0b series, now we have automatic digest-based image pulling in the agent side with a full integration with public/private Docker registries.

The next steps:
* [ ] Ability to snapshot a live container
* [ ] Ability to push the snapshotted image to a local registry

achimnol:
* If we allow dot-prefixed vfolders (e.g., `.local`), users will be able to keep their user-specific installation of Python packages across different compute sessions. #144 enables this.
 * For security, we are now moving forward to disallow root inside containers in all setups. In this sense, we need a new way to customize the docker images.
   - #171 allows importing an external Docker image with some constraints.
   - We could use [userspace image build tools](https://github.com/lablup/backend.ai-manager/pull/171#issuecomment-517541125) to allow rootless custom image build operation.

achimnol:
We have deployed a partially generalized solution as #171.
Currently our following approaches render the necessity of having container snapshots:
 * We don't allow root inside containers.
 * We let users install their own user-packages in the user directory, which is mounted from the host-side scratch directory.
 * We provide several ways to keep the user-specific content across different compute sessions by lablup/backend.ai#98 and lablup/backend.ai#99.

Closing.

PR #171, opened 2019-07-31
Image import

achimnol:
refs #119 

This PR adds a new API `POST /image/import` which converts an existing Docker image from one of the configured registries to a Backend.AI-compatible kernel image.

Currently we only support import of Docker images which contains a Python runtime (version 2.7, 3.5, 3.6, 3.7) and are compatible with [the Python's manylinux2010 ABI](https://www.python.org/dev/peps/pep-0571/). This includes NVIDIA GPU Cloud (NGC) images and most Debian/Ubuntu/CentOS-based data science images, with a notable exception of Alpine-based Docker images.
The importer kernel uses an auto-generated Dockerfile to convert the source image to a target Backend.AI kernel image with binary wheels built for the auto-detected Python version used by the source image. The wheel set contains dependencies for Jupyter/IPython kernels so that the imported image can be used with Jupyter service ports and via the query mode execution API.

See also: https://github.com/lablup/backend.ai-kernels/tree/master/importer


achimnol:
We need the followings:
 * Agents must be able to terminate the importer session once its job is completed.
   - This could be used for signalling completion of batch jobs in the ML pipeline.
 * We need to restrict custom code execution on the importer sessions. Let's add a "executable" flag to the `kernels` table.
   - This could be used for restricting interruption of batch jobs in the ML pipeline.

achimnol:
Batch job is separated as an issue: lablup/backend.ai#42

achimnol:
Let's consider using rootless userspace container build tools such as [kaniko](https://github.com/GoogleContainerTools/kaniko), [makisu](https://github.com/uber/makisu), and [buildah](https://buildah.io/).

achimnol:
Since lablup/backend.ai#42 is now implemented, we are ready to implement this.

PR #173, opened 2019-08-02
Add initial setup for changing RPC to Callosum's Peer-to-Peer RPC.

temirrr:
There are still many things to do and this commit is more of a skeleton for subsequent work.
I'm submitting this PR so that it could be seen which changes were introduced but apparently not intending to merge it into master by now.


temirrr:
I think one thing that can be not clear from the first time is why I instantiate Peer class twice in the RPCContext context manager. This is because, as we discussed with @achimnol , there has to be first connection which can communicate the **unique stream_key** for each manager-agent peer-to-peer communication. After that, we can open a peer class connecting to the stream with unique key.

PR #197, opened 2019-09-25
Consistent logging format

achimnol:
Let's define some rules for logging convention, particularly for API requests.


achimnol:
Also changed `gateway.etcd`'s configuration APIs and `gateway.manager`'s status update APIs to be superadmin only.

PR #202, opened 2019-10-29
Add support for handing service url template

kyujin-cho:
This PR adds new `extrainfo` endpoint to stream API.
Check lablup/backend.ai-agent#152 and lablup/backend.ai#54 for more details.

### DIFF
--- a/src/ai/backend/gateway/stream.py
+++ b/src/ai/backend/gateway/stream.py
@@ -424,32 +424,38 @@ async def stream_proxy(request: web.Request, params: Mapping[str, Any]) -> web.S
 @server_status_required(READ_ALLOWED)
 @auth_required
 async def get_stream_apps(request: web.Request) -> web.Response:
-    registry = request.app['registry']
     sess_id = request.match_info['sess_id']
     access_key = request['keypair']['access_key']
+    resp = []
 
-    result = await asyncio.shield(
-        registry.get_service_apps(sess_id, access_key, None))
-    if result['status'] == 'failed':
-        msg = f"Failed to launch get extra info of session {sess_id}: {result['error']}"
-        raise InternalServerError(msg)
-    return web.json_response(result['data'])
+    async with request.app['dbpool'].acquire() as conn, conn.begin():
+        query = (
+            sa.select([
+                kernels.c.service_ports,
+            ])
+            .select_from(kernels)
+            .where(
+                (kernels.c.sess_id == sess_id) &
+                (kernels.c.access_key == access_key)
+            ) 
+        )
+        result = await conn.execute(query)
+        row = await result.first()
+        for item in row['service_ports']:
+            response_dict = {
+                'name': item['name'],
+                'protocol': item['protocol'],
+                'ports': item['container_ports'],
+            }
+            if 'url_template' in item.keys():
+                response_dict['url_template'] = item['url_template']
+            if 'allowed_arguments' in item.keys():
+                response_dict['allowed_arguments'] = item['allowed_arguments']
+            if 'allowed_envs' in item.keys():
+                response_dict['allowed_envs'] = item['allowed_envs']
+            resp.append(response_dict)
 
-
-@server_status_required(READ_ALLOWED)
-@auth_required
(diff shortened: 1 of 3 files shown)


src/ai/backend/gateway/stream.py
```diff
@@ -573,6 +599,7 @@
     app['stream_execute_handlers'] = defaultdict(weakref.WeakSet)
     app['stream_proxy_handlers'] = defaultdict(weakref.WeakSet)
     app['stream_stdin_socks'] = defaultdict(weakref.WeakSet)
+    app['stream_extra_info'] = defaultdict(weakref.WeakSet)
```
achimnol: This seems not to be used elsewhere. Please remove it if so.

src/ai/backend/manager/registry.py
```diff
@@ -843,7 +843,17 @@ def __init__(self, config_server, dbpool,
             async with RPCContext(kernel['agent_addr'], None) as rpc:
                 coro = rpc.call.start_service(str(kernel['id']), service, opts)
                 if coro is None:
-                    log.warning('stat_service cancelled')
+                    log.warning('start_service cancelled')
+                    return None
+                return await coro
+
+    async def get_service_extra(self, sess_id, access_key, service):
```
achimnol: Rename to `get_service_ports()`.

src/ai/backend/gateway/stream.py
```diff
@@ -388,6 +390,11 @@
 
     try:
         opts: Mapping[str, Any] = {}
+        if params['arguments'] != None:
```
achimnol: Use `is not` instead of `!=`.
See https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none
TL;DR: PEP-8 recommends `is` operator for `None` comparison and it is much faster.

src/ai/backend/gateway/stream.py
```diff
@@ -322,6 +322,8 @@
     t.Dict({
         tx.AliasedKey(['app', 'service']): t.String,
         tx.AliasedKey(['port'], default=None): t.Null | t.Int[1024:65535],
+        tx.AliasedKey(['arguments'], default=None): t.Null | t.String, # {'-P': '12345'}
+        tx.AliasedKey(['envs'], default=None): t.Null | t.String # {'PASSWORD': '12345'}
```
achimnol: Insert an additional space before the comment sign (`#`) as PEP-8 recommends to use *two or more* spaces before line-ending comments.

src/ai/backend/gateway/stream.py
```diff
@@ -322,6 +322,8 @@
     t.Dict({
         tx.AliasedKey(['app', 'service']): t.String,
         tx.AliasedKey(['port'], default=None): t.Null | t.Int[1024:65535],
+        tx.AliasedKey(['arguments'], default=None): t.Null | t.String, # {'-P': '12345'}
```
achimnol: Please include more formal definition of the value format like:
```suggestion
        tx.AliasedKey(['arguments'], default=None): t.Null | t.String,  # stringified JSON, e.g., '{"-P": "12345"}'
                                                                        # For flag-style arguments, the value must be None.
```

src/ai/backend/gateway/stream.py
achimnol: Please include more formal definition of the value format like:
```suggestion
        tx.AliasedKey(['envs'], default=None): t.Null | t.String     # stringified JSON, e.g., '{"PASSWORD": "12345"}'
```

kyujin-cho:
Published new commit which contains:
- agent will now return information of services which are defined in local docker images, but not in etcd
  - Priority of each service definition sources will be: intrinsic > etcd > local docker image label
- agent's `service_info` field(inside `create_kernel` call's response data) now contains service port's template info by agent querying live service info right after kernel creation
- stream's `/apps` API will now fetch service info from psql; This means no RPC calls are made when querying service info
- `/apps/{app_name}` API is removed
- `get_service_apps` RPC call is removed, since we don't use it anymore

PR #205, opened 2019-11-18
Fix psql query incorrectly written when updates dict is empty

kyujin-cho:


achimnol:
This is an empty merge because it's already included in #206 and master.

PR #206, opened 2019-11-19
mount vFolder on a custom location

kyujin-cho:
This PR resolves lablup/backend.ai#69 . 
What has changed:
- New API version `v4.20191119` created, which reflects new `mount_map` property when creating kernel.
- New JSONB column `mount_map` added to `kernel` table on alembic.
- Kernel API now accepts `mount_map` property inside `creation_config`.

### DIFF
--- a/src/ai/backend/gateway/kernel.py
+++ b/src/ai/backend/gateway/kernel.py
@@ -89,6 +89,7 @@ creation_config_v2 = t.Dict({
 })
 creation_config_v3 = t.Dict({
     t.Key('mounts', default=None): t.Null | t.List(t.String),
+    t.Key('mount_map', default=None): t.Null | t.Mapping(t.String, t.String),
     t.Key('environ', default=None): t.Null | t.Mapping(t.String, t.String),
     tx.AliasedKey(['cluster_size', 'clusterSize'], default=None):
         t.Null | t.Int[1:],
--- a/src/ai/backend/gateway/kernel.py
+++ b/src/ai/backend/gateway/kernel.py
@@ -89,6 +89,17 @@ creation_config_v2 = t.Dict({
 })
 creation_config_v3 = t.Dict({
     t.Key('mounts', default=None): t.Null | t.List(t.String),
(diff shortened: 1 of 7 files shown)


src/ai/backend/gateway/kernel.py
```diff
@@ -275,7 +287,9 @@ def interrupt_wait(ctx: Any, event_name: str, agent_id: AgentId,
                 raise InvalidAPIParameters('Invalid group')
 
         api_version = request['api_version']
-        if (4, '20190315') <= api_version:
+        if (4, '20191119') <= api_version:
```
achimnol: Conventionally we keep the API version to the middle of triple months (one of 0315, 0615, 0915, or 1215).
Since we will make a breaking change for the naming of kernels and sessions in v5, let's make the version to `v5.20191215`.

src/ai/backend/gateway/kernel.py
```diff
@@ -98,6 +98,18 @@
     tx.AliasedKey(['resource_opts', 'resourceOpts'], default=None):
         t.Null | t.Mapping(t.String, t.Any),
 })
+creation_config_v4 = t.Dict({
+    t.Key('mounts', default=None): t.Null | t.List(t.String),
+    t.Key('mount_map', default=None): t.Null | t.Mapping(t.String, t.String),
```
achimnol: For Javascript clients, let's add an alias `mountMap` using `tx.AliasedKey`.

src/ai/backend/gateway/kernel.py
```diff
@@ -287,7 +287,7 @@ def interrupt_wait(ctx: Any, event_name: str, agent_id: AgentId,
                 raise InvalidAPIParameters('Invalid group')
 
         api_version = request['api_version']
-        if (4, '20191119') <= api_version:
+        if api_version[0] <= 5:
```
achimnol: Isn't the operands reversed?

kyujin-cho: :( ashamed of myself

PR #211, opened 2019-12-12
Keypair-specific SSH keypair

adrysn:
Current SSH keypair used to connect to a container is created per container-basis. That means, user has to change private key for SSH/SFTP client, everytime a container is destroyed and created. Furthermore, when user spawns multiple containers and wants to connect to them alternately, the user has to change private key every time they connect to another container.

User can create `.ssh` virtual folder and store public key in `authorized_keys` to use the same private key. However, this method is not possible when group virtual folder is the only allowed type.

To ease the pain in those cases, this pull request supports per-keypair SSH keypair. When new keypair is created, manager automatically generates a new SSH keypair and saves them in DB (`keypairs.c.ssh_public_key` and `keypairs.c.ssh_private_key`). This keypair is passed to the kernel creation step by `internal_data`. Then, the agent uses the information to mount `./ssh/authorized_keys` (stores public key) and `id_container` (private key).

Also, I added two API endpoints for getting and refreshing SSH keypair like below. There are no parameters for the endpoints.

```
GET  /auth/ssh-keypair
POST /auth/ssh-keypair
```

This pull-request is branched from 19.09 version.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,7 @@ install_requires =
     async_timeout~=3.0  # to avoid pip 10 resolver issue
     attrs>=18.0         # to avoid pip 10 resolver issue
     click>=7.0
+    cryptography>=2.8
     dataclasses; python_version<"3.7"
     graphene~=2.1.0
     Jinja2~=2.10.1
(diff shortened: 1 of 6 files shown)


achimnol:
For the refresh handler, let’s change the method to `PATCH`.

adrysn:
I've changed the method from POST to PATCH for  SSH key refresh handler.

PR #212, opened 2019-12-19
Scheduling plugin and DRF scheduler

achimnol:
This PR:
 * [x] Provides an abstract scheduler interface in two steps: `pick_session()` and `assign_agent()`
   - These are pure functions, but a scheduler implementation may keep contextual information for a single scaling group as instance variables and task-local information as contextvars.
 * [x] Allows installation of user-defined scheduler as plugins
 * [x] Adds an intrinsic scheduler that implements [the DRF algorithm](https://www.usenix.org/conference/nsdi11/dominant-resource-fairness-fair-allocation-multiple-resource-types).

The scheduling decision is made per-scaling-group.
If a session may run in multiple scaling groups, its scheduling is tried in all the requested (and allowed) scaling groups.


achimnol:
Remaining TODOs:
 * Make it tests runnable and add `tests/manager/test_scheduler.py`
 * Reimplement `reserve_agent` and `unreserve_agent` operations

PR #213, opened 2019-12-19
Add task template and bootstrap script feature when creating kernel

kyujin-cho:
This PR resolves issue lablup/backend.ai#70 and lablup/backend.ai#77.

# What has changed
## Bootstrap Script
Kernel Creation API now accepts `bootstrap_script` parameter. Value of `bootstrap_script` should be encoded to Base64. 
## Task Template support
- new `/task-template` API is added, which supports basic CRUD operation for task template.
- Kernel Creation API now accepts `template_id` parameter, which represents a valid task template UUID. 
  - If same parameters/configs are specified in both task template and Kernel Creation API body, manager will  use later one


src/ai/backend/gateway/template.py
```diff
@@ -0,0 +1,369 @@
...
+
+async def shutdown(app: web.Application) -> None:
+    pass
+
+
+def create_app(default_cors_options: CORSOptions) -> Tuple[web.Application, Iterable[WebMiddleware]]:
+    app = web.Application()
+    app.on_startup.append(init)
+    app.on_shutdown.append(shutdown)
+    app['api_versions'] = (1, 2, 3, 4)
```
achimnol: This API is a brand new one. Set this to `(4, 5)`.

PR #217, opened 2020-01-09
Implement basic multi container session

kyujin-cho:
This PR implements basic multi container session, as described in lablup/backend.ai#78 and lablup/backend.ai#70. 

## What has changed
* Update terminologies
  - "multi-container session" is also called "cluster session".
  - Now *session* indicates a user-perceived unit of computation jobs, which may contain one or more *kernels* (or *containers*, *nodes*) as units of independent computation workers
  - *master* container is renamed to *main* container.
* New `/template/cluster` endpoint, which provides CRUD operation for cluster templates
* New `/session/_/cluster` endpoint to create multi-container (a.k.a. cluster) session
  - Clients should specify per-role templates and their number of replicas.
* Update `/session` endpoint to a create multi-container session when:
  - The cluster size parameter is set larger than 1 but only one kernel configuration is specified.
* Two modes of operation
  - **single-node**: All containers that belongs the same session run on the same agent (the current default)
    - It is scheduled as a single chunk of resources.
  - **multi-node**: Each container may reside on different agent nodes (BETA)
    - Each node is scheduled separately.
* Update `/events/session` endpoint
  - Extended events 
    (`kernel_xxx` are still fired but you may see multiple occurrences of them for multi-container sessions, while `session_xxx` are fired only once per session)
    - `kernel_started` -> `session_started`
    - `kernel_terminated` -> `session_terminated`
    - `kernel_cancelled` -> `session_cancelled`
  - Renamed events 
    (`kernel_xxx` are no longer fired but only `session_xxx` are)
    - `kernel_success` -> `session_success`
    - `kernel_failure` -> `session_failure`
  - `/events/session` endpoint
    - It now has a `scope` parameter which may be `"*"`, `"session"`, `"kernel"` to filter above events.
    - It now acceps an optional session ID in the UUID format as `sessionId` parameter, which makes the API to ignore the `sessionName` parameter.
    - All `kernel_*` events now have explicit `kernelId` and `sessionId` fields.
    - All `session_*` events have an explicit `sessionId` field.
* Update SSH support
  - In all cluster sessions with two or more nodes, all nodes (conatiners) start with the SSH daemon running at the default 2200 service port, and a random-generated keypair is configured for password-less accesses to all of the cluster hostnames in the cluster.
  - This allows seamless execution of `horovodrun` and other distributed computing commands that utilizes SSH to initiate their work across multiple nodes.
* Idle timeout check
  - It is now done by the manager, because agents cannot update the "recent used" timestamp of containers running in other agents.
* All APIs returning the list of sessions/kernels now return the list of sessions.
  - The kernels belonged to each session must be queried separately, via GraphQL `compute_session { containers { ... } }` field.
  - The resource usage and statistics are now the sum of all of belonging kernels.
  - This allows existing console/UI and client SDKs to work seamlessly, with only the updates for the session event monitoring API.

## Information about cluster session
* Single-container sessions is also treated as a cluster session, but only with 1 `main` node
* Multi-container sessions can have any type of roles, although there are some restrictions:
  - Each role can be defined only once.
  - The `main` role is required for cluster.
  - Only one `main` node can be created.
  - When automatically replicated, non-`main` nodes have the `sub` role.
  - Each node is given a clsuter hostname, composed of "{role}{idx}", e.g., `main1`, `sub1`, `sub2` in a 3-node auto-replicated cluster.
* Service ports: only available in the `main` node, and users interacts with the main node only.
* Inter-node networking
  - single-node mode: uses the local Docker daemon's private bridge network
  - multi-node mode: uses the Docker Swarm's overlay network
  - In a cluster, containers may connect to other containers using their cluster hostnames, without explicit port mappings.
  - Containers cannot access to containers in other clusters.
* Inside all kernels, the following list of environment variables are provided:
  - `BACKENDAI_CLUSTER_SIZE`: the number of all nodes in the current cluster
  - `BACKENDAI_CLUSTER_MODE`: `"single-node"` or `"multi-node"`
  - `BACKENDAI_CLUSTER_REPLICAS`: a comma-separated list of colon-separated pairs of configured cluster roles and the number of replicas for each role
  - `BACKENDAI_CLUSTER_HOST`: the hostname of this node (composed of the role name and the index)
  - `BACKENDAI_CLUSTER_HOSTS`: a comma-separated list of all the hostnames in the current cluster
  - `BACKENDAI_CLUSTER_ROLE`: the name of the current node's role
  - `BACKENDAI_CLUSTER_IDX`: the 1-based index of the current node among the nodes in the same cluster role
  - `BACKENDAI_SESSION_ID`: the session ID
  - `BACKENDAI_KERNEL_ID`: the kernel (node) ID


achimnol:
Remaining TODOs:
* idle timeout check in the manager side
* restarting of multi-kernel sessions

PR #219, opened 2020-01-16
Make all vfolder operations to be async

achimnol:
This is a hotfix to keep all vfolder (i.e., NFS) operations async.

refs lablup/backend.ai#102


achimnol:
Tested working of all patched APIs.

PR #232, opened 2020-02-11
Collect container logs at container termination

kyujin-cho:
This PR resolves issue lablup/backend.ai#115.


src/ai/backend/manager/models/alembic/versions/51dddd79aa21_add_logs_column_on_kernel_table.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '51dddd79aa21'
+down_revision = '1e8531583e20'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('kernels', sa.Column('container_log', sa.String, nullable=True))
```
achimnol: Let's change this column type to [`sa.LargeBinary`](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.LargeBinary) to store the given log data as-is.
Since the log may contain non-Unicode characters and garbled outputs (due to mishandling of encodings in the user codes, etc.), it would be safer to use binary types to store them.
This will also allow appropriate conversion and streaming in the database backends if available.

src/ai/backend/manager/models/kernel.py
```diff
@@ -132,7 +132,7 @@ class KernelStatus(enum.Enum):
               server_default=SessionResult.UNDEFINED.name,
               nullable=False, index=True),
     sa.Column('internal_data', pgsql.JSONB(), nullable=True),
-
+    sa.Column('container_log', sa.String, nullable=True),
```
achimnol: Same here.

src/ai/backend/gateway/session.py
```diff
@@ -689,6 +689,16 @@ def interrupt_wait(ctx: Any, event_name: str, agent_id: AgentId,
     await app['registry'].handle_stats(agent_id, kern_stats)
 
 
+async def handle_kernel_log(app: web.Application, agent_id: AgentId, event_name: str,
+                            raw_kernel_id: str, log_str: str):
```
achimnol: As commented in the agent PR, let's use a separate Redis key to store/retrieve the log data as list of Redis strings.

src/ai/backend/gateway/session.py
```diff
@@ -700,6 +701,27 @@ def interrupt_wait(ctx: Any, event_name: str, agent_id: AgentId,
...
+    dbpool = app['dbpool']
+    connection: aioredis.Redis = await redis.connect_with_retries(
+                app['config']['redis']['addr'].as_sockaddr(),
+                db=REDIS_STREAM_DB,
+                password=(app['config']['redis']['password']
+                          if app['config']['redis']['password'] else None),
+                encoding=None,
+    )
+    logs = await redis.execute_with_retries(
+            lambda: connection.lrange(f'containerlog.{container_id}', 0, -1))
```
achimnol: Please split this into multiple calls of LPOP.
The reason to use Redis lists is *not* to block the connection while we are sending/receiving large data to/from Redis.

PR #235, opened 2020-02-13
Implement logs API

kyujin-cho:
This PR resolves lablup/backend.ai#110.

Ticket: OP#162


scripts/drop-user-related-columns.sql
```diff
@@ -0,0 +1,8 @@
+DELETE FROM domains;
+DELETE FROM groups;
+DELETE FROM scaling_groups;
+DELETE FROM sgroups_for_domains;
+DELETE FROM users;
+DELETE FROM association_groups_users;
+DELETE FROM keypair_resource_policies;
+DELETE FROM keypairs;
```
achimnol: `scripts/drop-user-related-columns.sql` does not match with what it actually does.
If it's added for your local debugging, please remove it.

src/ai/backend/gateway/logs.py
```diff
@@ -0,0 +1,203 @@
...
+        t.Key('source'): t.String,
+        t.Key('message'): t.String,
+        t.Key('context_lang'): t.String,
+        t.Key('context_env'): tx.JSONString,
+        t.Key('request_url', default=None): t.Null | t.String,
+        t.Key('request_status', default=None): t.Null | t.Int,
+        t.Key('traceback', default=None): t.Null | t.String
+    }
+))
+async def create(request: web.Request, params: Any) -> web.Response:
```
achimnol: It may not look important, but considering the nature of a logging system, it would be better to use names like "append".

src/ai/backend/gateway/logs.py
```diff
@@ -0,0 +1,203 @@
...
+                                 (error_logs.c.id == log_id)))
+
+        result = await conn.execute(query)
+        assert result.rowcount == 1
+
+        return web.json_response({'success': True}, status=200)
+
+
+async def init(app: web.Application) -> None:
+    pass
```
achimnol: We should add a periodic timer task which cleans up old-enought logs (e.g., appended before 3 months).

src/ai/backend/gateway/logs.py
```diff
@@ -0,0 +1,203 @@
...
+        result = await conn.execute(query)
+        assert result.rowcount == 1
+    return web.json_response(resp)
+
+
+@auth_required
+@server_status_required(READ_ALLOWED)
+@check_api_params(
+    t.Dict({
+        t.Key('mark_read', default='false'): t.Enum('true', 'True', 'false', 'False'),
```
achimnol: Let's use `t.ToBool()`.

src/ai/backend/manager/models/error_logs.py
```diff
@@ -0,0 +1,36 @@
...
+__all__ = [
+    'error_logs',
+    'LogSeverity'
+]
+
+
+class LogSeverity(str, enum.Enum):
+    CRITICAL = 'critical'
+    ERROR = 'error'
+    WARNING = 'warning'
```
achimnol: For future, let's add INFO, DEBUG levels as well.
(Don't forget to update the migration!)

src/ai/backend/gateway/logs.py
```diff
@@ -0,0 +1,203 @@
...
+            update = (sa.update(error_logs)
+                        .values(is_read=True)
+                        .where(error_logs.c.id.in_([x['log_id'] for x in resp['logs']])))
+            await conn.execute(update)
+        return web.json_response(resp, status=200)
+
+
+@auth_required
+@server_status_required(READ_ALLOWED)
+async def clear(request: web.Request) -> web.Response:
```
achimnol: "clear" may be misunderstood as "delete" for the first-time readers.
Let's rename this to "mark_cleared".

PR #238, opened 2020-02-18
Reserved words for vfolder

kyujin-cho:
This PR adds reserved words feature for vFolder.


achimnol:
Also, please add `.ssh` to the list of reserved words as well.

achimnol:
And for dotfiles that we automatically mount for all session by default (e.g., `.bashrc`), if the user provides them as user-specific dotfiles, the user-defined ones must precede.

kyujin-cho:
Ticket: OP#188

achimnol:
Also, please add some unit tests for those functions, as a practice.

PR #240, opened 2020-02-19
Prevent duplicate creation between dotfile and dot-prefix vFolder

kyujin-cho:
This PR adds checking procedure when adding vFolder or dotfile.


kyujin-cho:
Ticket: OP#189

src/ai/backend/gateway/userconfig.py
```diff
@@ -19,9 +19,9 @@
 from .utils import check_api_params, get_access_key_scopes
 
 from ..manager.models import (
-    keypairs, verify_dotfile_name
+    keypairs, vfolders, verify_dotfile_name, query_accessible_vfolders,
+    query_owned_dotfiles, MAXIMUM_DOTFILE_SIZE
```
achimnol: Please add the trailing comma.

PR #250, opened 2020-03-10
Support for forced termination of sessions

achimnol:
Let's add a programmatic, automated way that does what we do for manual agent recovery.


achimnol:
Internal ticket: OP#224

PR #257, opened 2020-03-19
Upgrade image import API

achimnol:
Upgrade the image import API to support the followings:

* Service-definition JSON files for custom service ports
  - The console UI should provide:
    - A syntax-highlighted JSON editor (it should generate a list of JSON files)
    - A set of templates such as exemplifying "jupyter" and "vnc-web".
* A new runtime-type "app"
  - When chosen, the imported image is almost unmodified but
    only labels and service-definitions are added.
* Auto-fill the labels if the source image is already Backend.AI-enabled image.
* Advanced mode
  - The user may write a part of Dockerfile directly (e.g., `RUN apt-get install ...`),
    with a grayed-out unmodifiable parts for Backend.AI-specific labels.

### DIFF
--- a/src/ai/backend/gateway/image.py
+++ b/src/ai/backend/gateway/image.py
@@ -186,12 +186,12 @@ async def get_import_image_form(request: web.Request) -> web.Response:
                     {
                         'name': 'runtimeType',
                         'type': 'choice',
-                        'choices': ['python'],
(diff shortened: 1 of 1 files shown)


achimnol:
Internal ticket: OP#277

achimnol:
This is replaced with an internal project.

PR #260, opened 2020-04-02
fix: allow group level allowed hosts for user

adrysn:
This pull request resolves an issue that a user folder cannot be created if `allowed_vfolder_hosts` is set only in the group-level (namely, host is not set in domain-/keypair-policy-level).

The cause of the problem is that the `allowed_vfolder_hosts` specified at the group level is not used when creating a user folder. Fixed to call `get_allowed_vfolder_hosts_by_user` instead of `get_allowed_vfolder_hosts_by_group` if there is no `group_id` parameter when creating a folder (meaning the request tries to create a user folder).


adrysn:
Oops! I have to create a pull request from master. My mistake. I'll close this PR.

PR #265, opened 2020-04-13
feat: extend vfolder to have innate permission and usage mode

adrysn:
This pull-request adds `permission` and `usage_mode` columns to `vfolders` table to extend the permission control and to deal with various purposes on how to use those folders. By introducing those fields, we can categorize vfolders in terms of purpose of use (general, model storage, public data storage, ...), and create a read-only group vfolder (which was not possible previously).

* DB Changes
  - Added `usage_mode` field to `vfolders`.
    - One of `general`, `data` or `model`.
    - Previous folders defaults to `general`.
  - Added `permission` field to `vfolders`.
    - Virtual folders now have innate permission field, and that should be considered in group folder operations. For example, a group folder can be shared as read-only by setting this field to `ro`.
    - Invited folders have it's own permission table. It overrides vfolder's own permission field.
    - `permission`: one of `ro`, `rw`, or `wd`
    - Alembic migration defaults it to `wd` and `rw` for user and group folders, respectively.
    - Regardless of the value of permission, folder owners can still write to and delete vfolder itself.
  - Added `ownership_type` field to `vfolders`.
    - One of `user` or `group`
    - To easily recognize a virtual folder is whether user or group type. `type` was too broad name.
    - `user` and `group` columns are still be kept.
    - Alembic migration automatically set the value by looking at the existence of `user` or `group`.
  - Constraints added
    - `ownership_type` value should match with existence of `user` or `group` column (`ck_vfolders_ownership_type_match_with_user_or_group`).
    - `user` and `group` columns should exclusively exist (`ck_vfolders_either_one_of_user_or_group`).
* API and/or logic changes
  - `query_allowed_vfolders` return new fields: `usage_mode`, `ownership_type`, and `permission`.
  - Updated `vfolder_permission_required` for group folders to consider new `permission` field.
  - Create vfolder API handler accepts `usage_mode` and `permission` parameter.
  - Allow users who have `RW_DELETE` perm can delete vfolder.
* Permission changes
  - Let superadmins can create group virtual folders.
  - Enable tus upload for `READ-WRITE` vfolder (from original `RW-DELETE`).
    - I see no reason to have delete permission to upload a file.
  - Enable vfolder file deletion for users who have `READ-WRITE` permission for that folder.
    - Original: `RW-DELETE`.
    - User can delete files from `READ-WRITE` vfolders anyway inside a session. We need to be consistent.


src/ai/backend/gateway/vfolder.py
```diff
@@ -178,6 +201,8 @@ def get_folder_hostpath(row: VFolderRow, app):
     t.Dict({
         t.Key('name'): tx.Slug(allow_dot=True),
         t.Key('host', default=None) >> 'folder_host': t.String | t.Null,
+        t.Key('usage_mode', default='general'): t.String | t.Null,
+        t.Key('permission', default='rw'): t.String | t.Null,
```
achimnol, line 205: How about using `tx.Enum` (`tx` is an aliased import of `ai.backend.common.validators`) to validate earlier?

src/ai/backend/manager/models/alembic/versions/529113b08c2c_add_vfolder_type_column.py
```diff
@@ -0,0 +1,127 @@
...
+        'ownership_type_match_with_user_or_group',
+        'vfolders',
+        '(ownership_type = \'user\' AND "user" IS NOT NULL) OR '
+        '(ownership_type = \'group\' AND "group" IS NOT NULL)'
+    )
+    op.create_check_constraint(
+        'either_one_of_user_or_group',
+        'vfolders',
+        '("user" IS NULL AND "group" IS NOT NULL) OR ("user" IS NOT NULL AND "group" IS NULL)'
+    )
```
achimnol, line 110: :+1:

src/ai/backend/manager/models/vfolder.py
```diff
@@ -82,8 +109,22 @@ class VFolderInvitationState(str, enum.Enum):
     sa.Column('creator', sa.String(length=128), nullable=True),
     # For unmanaged vFolder only.
     sa.Column('unmanaged_path', sa.String(length=512), nullable=True),
+    sa.Column('ownership_type', EnumValueType(VFolderOwnershipType),
+              default=VFolderOwnershipType.USER, nullable=False),
     sa.Column('user', GUID, sa.ForeignKey('users.uuid'), nullable=True),
     sa.Column('group', GUID, sa.ForeignKey('groups.id'), nullable=True),
+
+    # Note: Custom constraints should be manually defined during alembic migration.
+    #       This is just for reference.
```
achimnol, line 118: Explicit SQLAlchemy declaration is required for "oneshot" schema creation, not only for reference. :wink:

src/ai/backend/manager/models/vfolder.py
```diff
@@ -108,8 +149,8 @@ class VFolderInvitationState(str, enum.Enum):
     sa.Column('invitee', sa.String(length=256), nullable=False),  # email
     sa.Column('state', EnumValueType(VFolderInvitationState),
               default=VFolderInvitationState.PENDING),
-    sa.Column('created_at', sa.DateTime(timezone=True),
-              server_default=sa.func.now()),
+    sa.Column('created_at', sa.DateTime(timezone=True), server_default=sa.func.now()),
+    sa.Column('modified_at', sa.DateTime(timezone=True), nullable=True),
```
achimnol, line 153: It seems that there is no place to update this field.

adrysn: Make `modified_at` updated whenever mutation performed. The value will also be returned when querying virtual folders.

src/ai/backend/gateway/vfolder.py
adrysn: Fixed!

adrysn:
> * There seems no change that propagates the combination result of two permissions (innate & invited) to the agent's `create_kernel` RPC call. Please update the kernel creation API handler as well.

I think there is no need to change session creation logic since it queries to-be-mounted vfolders by calling `query_accessible_vfolders`, which already [overrides innate permission with invited one for shared vfolders](https://github.com/lablup/backend.ai-manager/pull/265/files#diff-386ccdfbb833087a80551588dc8bc933R270). For user's own vfolders and group vfolders, `query_accessible_vfolders` just uses innate permission.

For testing, I created following vfolders and invited every user-type vfolders to a user account (`dfa9da...`).
image

`group-readonly` and `group-readwrite` means the vfolders have innate permission of read-only and read-write, respectively. For user-type vfolders, the first word is innate permission and the permission after `-invited-` represents invited permission for the user account. For example, `readonly-invited-rw` vfolder is created with read-only innate permission, but shared with read-write permission for the user account.

After logging in with the user account, I created a session which mounts all vfolders shown above and tried to touch a file in each directory. And the result seems to be okay.
image

src/ai/backend/manager/models/vfolder.py
adrysn: Custom constraints will be set by migrating from existing DB, but I'm having trouble to set them directly in the table definition. I'll keep trying to resolve this. Any ideas are welcome.

adrysn: I checked that those constraints are placed successfully by executing `install-dev.sh` script. I think this problem is resolved.

PR #268, opened 2020-04-21
refactor: return NaN if group-level resource status is not allowed

adrysn:
* Return NaN when group-level resource status is not allowed to the client.
* Etcd config naming: `allow_group_total` -> `group_resource_visibility`.


src/ai/backend/gateway/resource.py
```diff
@@ -237,12 +237,12 @@ async def check_presets(request: web.Request, params: Any) -> web.Response:
...
         # Return group stats with zeros if not allowed.
-        allow_group_total = \
-            await request.app['registry'].config_server.get('config/api/resources/allow-group-total')
-        if allow_group_total != '':
-            group_limits = ResourceSlot({k: Decimal(0) for k in known_slot_types.keys()})
-            group_occupied = ResourceSlot({k: Decimal(0) for k in known_slot_types.keys()})
-            group_remaining = ResourceSlot({k: Decimal(0) for k in known_slot_types.keys()})
+        group_resource_visibility = await request.app['registry'].config_server.get(
+                'config/api/resources/group_resource_visibility')
+        if group_resource_visibility != '':
```
achimnol, line 242: Since `config_server.get()` returns `None` if the key does not exist, this condition shows the group resources *only when* the key is explicitly set to an empty string and otherwise hides them.
Is this the intended behavior? If I were you, I would use an explicit string constants like `"true"`, `"false"` with `trafaret.ToBool()`.
Please clarify here.

src/ai/backend/gateway/etcd.py
```diff
@@ -45,7 +45,7 @@
      + api
        - allow-origins: "*"
        + resources
-         - allow-group-total: ""  # return total group resource limits in check-presets
+         - group_resource_visibility: ""  # return group resource status in check-presets
```
achimnol, line 48: Thanks for the naming update.

src/ai/backend/gateway/resource.py
adrysn: Yes, it was intended to check the existence of the value. I agree that your suggestion reads more clear, so I updated the value should have explicit values such as `'true'`, `'false'`, `True`, `False`, etc. `None` value will also be translated as `False` (default behavior).

PR #276, opened 2020-05-04
fix: set correct scaling group name during session scheduling

adrysn:


adrysn:
I added a news fragment!

PR #278, opened 2020-05-11
Download a directory from vfolder as an archive

adrysn:
This PR adds a feature to serve a directory inside a virtual folder as an
archive to a client. To create a zip stream, I used
[zipstream-new](https://github.com/arjan-s/python-zipstream).

This PR closes lablup/backend.ai#180


src/ai/backend/gateway/vfolder.py
```diff
@@ -910,6 +913,74 @@ def _do_ops():
...
 
 
+async def download_directory_as_archive(request: web.Request,
+                                        file_path: Path,
+                                        zip_filename: str = None) -> web.StreamResponse:
+    """Serve a directory as a zip archive on the fly."""
+    class _Sentinel:
+        pass
+
+    _sentinel = _Sentinel()
```
achimnol, line 923: Please use `ai.backend.manager.types.Sentinel` (read the comment there about how to use it) for consistency with other codes. This allows use of strict type annotations such as `janus.Queue[Union[Sentinel, ItemType]]` and makes the conditional statement for the sentinel value to represent specialization of the item type into `ItemType` solely.

achimnol: Note that, using the same pattern, we can distinguish sentinels for different contexts, such as `QueueSentinel`, `IOSentinel`, `RXSentinel`, etc.

src/ai/backend/gateway/vfolder.py
```diff
@@ -910,6 +913,74 @@ def _do_ops():
...
+                while True:
+                    if not q.full():
+                        q.put(item)
+                        break
+                    time.sleep(0.01)
+            while True:
+                if not q.full():
+                    q.put(_sentinel)
+                    break
+                time.sleep(0.01)
```
achimnol, line 938: Since putting the sentinel object has negligible I/O and CPU overheads, I think we can just `q.put(_sentinel)` here.

src/ai/backend/gateway/vfolder.py
```diff
@@ -910,6 +913,74 @@ def _do_ops():
...
+
+    def _iter2aiter(iter):
+        """Iterable to async iterable"""
+        def _consume(loop, iter, q):
+            for item in iter:
+                while True:
+                    if not q.full():
+                        q.put(item)
+                        break
+                    time.sleep(0.01)
```
achimnol, line 933: Just use `q.put(item)` and it will block for you when the queue is full and resume the operation automatically when the consumer makes a free slot. While blocked, `iter` is also paused. Your pattern is only required for `put_nowait()` and currently I don't see any reason to stick with that.
We are already doing the flow-control by limiting the maximum queue size.

src/ai/backend/gateway/vfolder.py
```diff
@@ -910,6 +913,74 @@ def _do_ops():
...
+            finally:
+                q.close()
+                await q.wait_closed()
+
+        return _aiter()
+
+    if zip_filename is None:
+        zip_filename = file_path.name + '.zip'
+    zf = zipstream.ZipFile(compression=zipstream.ZIP_DEFLATED)
+    # zf = zipstream.ZipFile()
```
achimnol, line 961: Please clean up.

src/ai/backend/gateway/vfolder.py
```diff
@@ -968,7 +1039,7 @@ async def download_single(request: web.Request, params: Any, row: VFolderRow) ->
     except (ValueError, FileNotFoundError):
         raise GenericNotFound('The file is not found.')
     if not file_path.is_file():
-        raise InvalidAPIParameters('The file is not a regular file.')
+        return await download_directory_as_archive(request, file_path)
```
achimnol, line 1042: I think.. it's better *not* to implicitly change the mode of behavior, unless we have an explicit parameter to order so.
Let's add a new API endpoint for downloading directories.

src/ai/backend/gateway/vfolder.py
```diff
@@ -1048,7 +1119,7 @@ async def download_with_token(request) -> web.StreamResponse:
     except (ValueError, FileNotFoundError):
         raise InvalidAPIParameters('The file is not found.')
     if not file_path.is_file():
-        raise InvalidAPIParameters('The file is not a regular file.')
+        return await download_directory_as_archive(request, file_path)
```
achimnol, line 1122: Same here.

src/ai/backend/gateway/vfolder.py
adrysn: Replaced `_Sentinel` class with `ai.backend.manager.types.Sentinel`'s token.

src/ai/backend/gateway/vfolder.py
adrysn: Reduce above code block with just `q.put(_sentinel)`.

src/ai/backend/gateway/vfolder.py
adrysn: Reduced above code block with just `q.put(item)`.

src/ai/backend/gateway/vfolder.py
adrysn: Removed commented line.

src/ai/backend/gateway/vfolder.py
adrysn: I added `archive: boolean` parameter for downloading directory as an archive.

src/ai/backend/gateway/vfolder.py
adrysn: I added `archive: boolean` parameter for downloading directory as an archive.

PR #290, opened 2020-06-19
feat: option to reserve batch session

adrysn:
Add option to reserve batch session during session creation.

* `starts_at` option can only be effective with `BATCH`-type session. Otherwise, it will be ignored.
* `starts_at` parameter can either be an ISO8601 string or human-friendly timedelta string such as `30min`, `1h 20m`, etc. If unit is omitted, seconds are the default.
* `starts_at` parameter will be converted into Python's `datetime` object and stored in `kernels.c.starts_at`, which is a new column.
* Predicate `check_reserved_batch_session` is newly introduced. It will check if batch session's `starts_at` is greater than `now`, and if so, stop session creation.
* Note that it is not guaranteed a session to be started at the exact `starts_at` time. It would rather mean that the session will not be started until the `starts_at` time.

Related: lablup/backend.ai#130, OP#101


adrysn:
> Overall LGTM, but I'd like to request two changes:
> 
> 1. Make the API parameter & DB column naming more explicit.
>    Let's rename `reserve` to `starts_at`.
>    In the future, we might have a separate `reservation` table which will let the scheduler to automatically enqueue new sessions with pre-calculated "starts_at" fields.
> 2. Please add datetime-comparison test cases with multiple timezones compared to each other, by mocking the predicate arguments.

1. Renamed `reserve` and `reserved_at` to `starts_at` both from API parameter and DB column.
2. Added [a test case](https://github.com/lablup/backend.ai-manager/pull/290/commits/dac1796e594fecff898eb4187c608582268ca88c) that compares multiple timezones.

PR #292, opened 2020-06-27
fix: Ensure kernel IDs for registry operations

achimnol:


achimnol:
Running load runner with 300 concurrent sessions works good, with only one TERMINATING session remaining, which was cleaned up using (non-forced) `backend.ai rm` command.

PR #300, opened 2020-07-24
fix: add context for the call to from_row in simple_db_mutate_returning_item

adrysn:
This PR adds `context` for the call to `from_row` in `simple_db_mutate_returning_item`.
* During creation of resource preset which uses `simple_db_mutate_returning_item`, for example, it raises error stating `unexpected error: from_row() missing 1 required positional argument: 'row'`
* Minor un-related fix: return user's status in authorize API. This will be needed by client-side to determine the user's status and show better error texts when user fails to login.

### DIFF
--- a/changes/300.fix
+++ b/changes/300.fix
@@ -1,0 +1,1 @@
+Adds context for the call to from_row in simple_db_mutate_returning_item.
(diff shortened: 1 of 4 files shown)


achimnol:
LGTM! Thanks for finding this out.

PR #302, opened 2020-07-27
feat: add purge mutations for users, groups, and domains

adrysn:
This PR adds `purge` mutations to users, groups, and domains. Those three models did not expose delete operation until now (there is delete operation, but actually what it does is to just inactivate objects) since we thought it is too dangerous. However, there are requests for this feature from customers, and we decided to expose the API by `purge` mutation.

* Add `purge` mutation for domains.
  - To purge a domain, there should be no bound groups and users.
* Add `purge` mutation for groups.
  - There is no condition to purge a group since group is not an essential ingredient for users. When a group is deleted, users belong to it just lost access to that group, but their account still functions well if there are in other groups.
  - Group's related virtual folders will also be purged together.
* Add `purge` mutation for users.
  - This will completely delete user's virtual folders, kernels, and keypairs.
  - Virtual folders shared with other users will not be deleted by default, and their ownership is migrated to the requested admin.
  - If `purge_shared_vfolders` parameter is delivered, shared virtual folders will also be deleted.

Related: OP#651.

### DIFF
--- a/src/ai/backend/manager/models/domain.py
+++ b/src/ai/backend/manager/models/domain.py
@@ -201,9 +201,9 @@ class ModifyDomain(graphene.Mutation):
 
 
 class DeleteDomain(graphene.Mutation):
-    '''
+    """
     Instead of deleting the domain, just mark it as inactive.
-    '''
+    """
     allowed_roles = (UserRole.SUPERADMIN,)
 
     class Arguments:
@@ -223,12 +223,12 @@ class DeleteDomain(graphene.Mutation):
 
 
 class PurgeDomain(graphene.Mutation):
-    '''
+    """
     Completely delete domain from DB.
 
     Domain-bound kernels will also be all deleted.
     To purge domain, there should be no users and groups in the target domain.
-    '''
+    """
     allowed_roles = (UserRole.SUPERADMIN,)
 
     class Arguments:
--- a/src/ai/backend/manager/models/domain.py
+++ b/src/ai/backend/manager/models/domain.py
@@ -196,7 +196,9 @@ class ModifyDomain(graphene.Mutation):
 
 
 class DeleteDomain(graphene.Mutation):
-
+    '''
+    Instead of deleting the domain, just mark it as inactive.
+    '''
     allowed_roles = (UserRole.SUPERADMIN,)
 
     class Arguments:
@@ -212,4 +214,32 @@ class DeleteDomain(graphene.Mutation):
             .values(is_active=False)
             .where(domains.c.name == name)
         )
+        return await simple_db_mutate(cls, info.context, query)
+
+
+class PurgeDomain(graphene.Mutation):
+    '''
+    Completely delete domain from DB.
+
+    To purge domain, there should be no users in the target domain.
+    '''
+    allowed_roles = (UserRole.SUPERADMIN,)
+
+    class Arguments:
+        name = graphene.String(required=True)
+
+    ok = graphene.Boolean()
+    msg = graphene.String()
+
+    @classmethod
+    async def mutate(cls, root, info, name):
+        from ai.backend.manager.models import  users
+        async with info.context['dbpool'].acquire() as conn:
+            query = (
+                sa.select([sa.func.count()])
+                .where(users.c.domain_name == name)
+            )
+            user_count = await conn.scalar(query)
+            assert user_count == 0, 'There are users bound to the domain. Remove users first.'
+        query = domains.delete().where(domains.c.name == name)
         return await simple_db_mutate(cls, info.context, query)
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/models/domain.py
```diff
@@ -17,6 +20,8 @@
 from .scaling_group import ScalingGroup
 from .user import UserRole
 
+log = BraceStyleAdapter(logging.getLogger('ai.backend.gateway.admin'))
```
achimnol, line 23: Usually it's sufficient to use `__file__`. Is there specific reasons to manually set this as the path for a different module?

src/ai/backend/manager/models/domain.py
```diff
@@ -196,7 +201,9 @@ async def mutate(cls, root, info, name, props):
 
 
 class DeleteDomain(graphene.Mutation):
-
+    '''
```
achimnol, line 204: For new codes, let's use double quotes for docstrings as PEP-8 and PEP-257 suggests.

src/ai/backend/manager/models/group.py
```diff
@@ -220,35 +229,26 @@ class Arguments:
...
-                    return cls(ok=True, msg='success', group=o)
-                else:
-                    return cls(ok=False, msg='failed to create group', group=None)
-            except (pg.IntegrityError, sa.exc.IntegrityError) as e:
-                return cls(ok=False, msg=f'integrity error: {e}', group=None)
-            except (asyncio.CancelledError, asyncio.TimeoutError):
-                raise
-            except Exception as e:
-                return cls(ok=False, msg=f'unexpected error: {e}', group=None)
+        assert _rx_slug.search(name) is not None, 'invalid name format. slug format required.'
```
achimnol, line 232: Let's change assert into an if statement with explicit raising of exceptions.
(If Python is built with optimization flags, assert statements may be ignored)

src/ai/backend/manager/models/domain.py
```diff
@@ -213,3 +220,64 @@ async def mutate(cls, root, info, name):
...
+
+        :param conn: DB connection
+        :param domain_name: domain's name to delete kernels
+
+        :return: number of deleted rows
+        """
+        from . import kernels
+        query = (
+            kernels.delete()
+            .where(kernels.c.domain_name == domain_name)
```
achimnol, line 278: I think we need to block this operation if there are non-terminated kernels in the given domain because the actual containers will become orphans and no longer tracked by Backend.AI, meaning resource leaks.

src/ai/backend/manager/models/group.py
```diff
@@ -336,19 +366,76 @@ class Arguments:
...
-            except Exception as e:
-                return cls(ok=False, msg=f'unexpected error: {e}')
+                loop = current_loop()
+                await loop.run_in_executor(None, lambda: shutil.rmtree(folder_path))  # type: ignore
+            except IOError:
+                pass
+        query = (
+            vfolders.delete()
+            .where(vfolders.c.group == group_id)
+        )
```
achimnol, line 413: We need to block this operation if there are vfolders mounted onto containers.

src/ai/backend/manager/models/group.py
```diff
@@ -336,19 +366,76 @@ class Arguments:
...
+        :param conn: DB connection
+        :param group_id: group's UUID to delete kernels
+
+        :return: number of deleted rows
+        """
+        from . import kernels
+        query = (
+            kernels.delete()
+            .where(kernels.c.group_id == group_id)
+        )
```
achimnol, line 437: We need to block this operation if there are non-terminated kernels.

src/ai/backend/manager/models/user.py
```diff
@@ -666,6 +679,246 @@ async def mutate(cls, root, info, email):
...
+
+                if not purge_shared_vfolders:
+                    await cls.migrate_shared_vfolders(
+                        conn,
+                        deleted_user_uuid=user_uuid,
+                        target_user_uuid=info.context['user']['uuid'],
+                        target_user_email=info.context['user']['email'],
+                    )
+                await cls.delete_vfolders(conn, user_uuid, info.context['config_server'])
+                await cls.delete_kernels(conn, user_uuid)
```
achimnol, line 727: We need to block these operations if there are non-terminated kernels or vfolders mounted onto running containers.

src/ai/backend/manager/models/domain.py
adrysn: I replaced single quotes to double quotes.

src/ai/backend/manager/models/group.py
adrysn: Replaced assert into explicit exceptions.

src/ai/backend/manager/models/domain.py
adrysn: Replaced it with `__file__`.

PR #305, opened 2020-07-27
Add a mini-language parser for generic GQL list filtering

Jaymel:
Defined a method to change the query statement to sql statement.


achimnol:
While I'm reviewing, please sign up the CLA and fix up the style issues indicated by lint checkers.

src/ai/backend/manager/models/minilang/queryfilter.py
```diff
@@ -0,0 +1,186 @@
...
+            'ne': '!='
+        }
+        self._JSON_PARSER = Lark(
+            Query2sql.__JSON_GRAMMER, parser='lalr', maybe_placeholders=False)
+
+    def query2tree(self, query):
+        if not checkBracket(query):
+            raise ValueError(r'''
+            Check the brackets in the query string.
+            ''')
```
achimnol, line 54: This is unncessary because the grammar already performs matching.
Test it by yourself.

src/ai/backend/manager/models/minilang/queryfilter.py
```diff
@@ -0,0 +1,186 @@
...
+            Check the quotes in the value string.
+            If you want to write quotes in the value string, use ' instead of ".
+            ''')
+        if not checkEscapeString(query):
+            raise ValueError(r'''
+            Be careful with the use of Escape string.
+            Escape string can only be used within value string.
+            Double Quotes(") can only be used to represent value string type.
+            If you want to write quotes in the value string, use ' instead of ".
+            ''')
```
achimnol, line 66: My expectation is that we should be able to have `\"` in `"`-strings and `\'` in `'`-strings, not only `"` in `'` strings and `'` in `"`-strings.

tests/manager/test_queryfilter.py
```diff
@@ -0,0 +1,135 @@
...
+
+Base.metadata.create_all(engine)
+session = Session()
+
+session.add_all([
+    User('tester', 'tester1', 30),
+    User('test\'er', 'tester2', 30),
+    User('test\'er ♪', 'tester3', 20)])
+
+session.commit()
```
achimnol, line 79: Put these as a fixture, instead of module-level codes, so that it is executed only when tests that actually requires the dummy db, and not when the test framework imports this module.

src/ai/backend/manager/models/minilang/queryfilter.py
```diff
@@ -0,0 +1,186 @@
...
+    def tree2where(self, tree):
+        filter_stack = []
+        count_stack = []
+        bracket_stack = []
+        where = ""
+        if tree.data != "filter":
+            tree = self.tree_partition(
+                tree, ret_table=False, ret_filter=True, ret_select=False)
+        tree = list(tree.iter_subtrees_topdown())
+        for i in range(len(tree)):
```
achimnol, line 116: How about using `enumerate()` to iterate over the index *and* the item with a single `for` statement?

src/ai/backend/manager/models/minilang/queryfilter.py
```diff
@@ -0,0 +1,186 @@
...
+                    where += token + " "
+                    if tree[i - 1].data != "pair":
+                        while bracket_stack:
+                            where = where[:-1]
+                            where += bracket_stack.pop()
+                    if filter_stack:
+                        where += self._FILTERS[filter_stack.pop()] + " "
+        if where[0] == "(" and where[-2] == ")":
+            where = where[1:-1]
+        return where[:-1]
```
achimnol, line 145: I think it would be better to generate the column object representation directly instead of generating the `where` string. Let's think about how we could start with a root expression, maybe `sa.sql.expression.true`.

tests/manager/test_queryfilter.py
```diff
@@ -0,0 +1,135 @@
...
+    assert where_clause == "(name = \"tester\" OR name = \"lablup\") AND age > 20"
+
+
+def test_deep_tree2select():
+    tree = q2s.query2tree(test_deep_json)
+    select_clause = q2s.tree2select(tree)
+    assert select_clause == "name, age"
+
+
+def test_escape_filter():
```
achimnol, line 118: Rename this following `escape_filter` renaming.

achimnol (approved):
Overall, LGTM about fulfilling my functional requirements.
Since your internship has finished, I will do some style/code clean up.

achimnol:
Now we need to use `ai.backend.manager.models.minilang.queryfilter.QueryFilterParser` in all paginated list GQL query resolver functions under `ai.backend.manager.models`.

achimnol:
I will file a new issue for the application of `QueryFilterParser`, after merging this.

PR #306, opened 2020-07-28
fix: prevent purging a user, group, and domain if bound resources are being used

adrysn:
Followup of https://github.com/lablup/backend.ai-manager/pull/302.

* Prevent purging a user when the user's virtual folders are mounted to some active sessions and/or the user has running sessions.
* Prevent purging a group when the group's virtual folders are mounted to some active sessions and/or the group has active sessions.
* Prevent purging a domain when it has active sessions.

### DIFF
--- a/src/ai/backend/manager/models/domain.py
+++ b/src/ai/backend/manager/models/domain.py
@@ -307,4 +307,4 @@ class PurgeDomain(graphene.Mutation):
                    (kernels.c.status.in_(AGENT_RESOURCE_OCCUPYING_KERNEL_STATUSES)))
         )
         active_kernel_count = await conn.scalar(query)
-        return True if active_kernel_count > 0 else False
+        return (active_kernel_count > 0)
--- a/src/ai/backend/manager/models/domain.py
+++ b/src/ai/backend/manager/models/domain.py
@@ -241,7 +241,7 @@ class PurgeDomain(graphene.Mutation):
         from . import users, groups
         async with info.context['dbpool'].acquire() as conn:
             if await cls.domain_has_active_kernels(conn, name):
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/models/domain.py
```diff
@@ -282,3 +284,27 @@ async def delete_kernels(
...
+        """
+        from . import kernels, AGENT_RESOURCE_OCCUPYING_KERNEL_STATUSES
+        query = (
+            sa.select([sa.func.count()])
+            .select_from(kernels)
+            .where((kernels.c.domain_name == domain_name) &
+                   (kernels.c.status.in_(AGENT_RESOURCE_OCCUPYING_KERNEL_STATUSES)))
+        )
+        active_kernel_count = await conn.scalar(query)
+        return True if active_kernel_count > 0 else False
```
achimnol, line 310: This may look more explicit, but I think it's better to just rely on the language semantics:
```suggestion
        return (active_kernel_count > 0)
```

src/ai/backend/manager/models/user.py
```diff
@@ -872,6 +878,68 @@ async def delete_vfolders(
...
+        """
+        from . import kernels, AGENT_RESOURCE_OCCUPYING_KERNEL_STATUSES
+        query = (
+            sa.select([sa.func.count()])
+            .select_from(kernels)
+            .where((kernels.c.user_uuid == user_uuid) &
+                   (kernels.c.status.in_(AGENT_RESOURCE_OCCUPYING_KERNEL_STATUSES)))
+        )
+        active_kernel_count = await conn.scalar(query)
+        return True if active_kernel_count > 0 else False
```
achimnol, line 941: Same here.
```suggestion
        return (active_kernel_count > 0)
```

PR #310, opened 2020-08-05
fix: vfolder.py integer type cast to parameter size

leksikov:
Despite the tus client sends file size parameter as integer. The server encodes parameter file size in jwt as a string. This will ensure that the parameter when it is used is integer.


achimnol:
Hm... the timeline protection bot seems to have stopped working, which checks the existence of news fragments in the `changes` directory. 🤔

PR #313, opened 2020-08-20
feat: include user information in keypair gql query

adrysn:
Some customers want to get some of users' information, such as full_name, when querying keypairs. They want to distinguish the owner of keypairs by user's full name, not by email (which they cannot remember).

So, This PR adds additional user fields, `email`, and `full_name`, in query keypair.

Related: https://github.com/lablup/backend.ai-client-py/pull/122.
Internal ticket: OP#672.


achimnol (approved):
LGTM!

PR #316, opened 2020-08-26
fix: regression of error logs API

achimnol:


achimnol:
Please add the new fragment here as well! (I cannot request changes as a review because I'm the first author of this PR...)

PR #319, opened 2020-08-31
script migrate vfolder structure

leksikov:
Script to migrate the vfolder vroot/local structure.

Example usage: python migrate-vfolder-structure.py /Users/name/vroot/local/ /Users/name/vroot/vfs/

This script will automatically create new folder structure and move files.

### DIFF
--- a/scripts/migrate-vfolder-structure.py
+++ b/scripts/migrate-vfolder-structure.py
@@ -8,33 +8,26 @@ def main():
     print(input_path, output_path)
     subfolders = [x for x in input_path.iterdir() if x.is_dir()]
 
-    print("Subfolders ", subfolders)
+    print("Number of folders ", len(subfolders))
+    try:
 
-    for folder in subfolders:
-        folder = str(folder)
-        folder = folder.split("/")[-1]
-        header_dir = Path(folder[0:2])
-        second_lvl_dir = Path(folder[2:4])
-        rest_dir = Path(folder[4:])
+        for folder in subfolders:
+            folder = str(folder)
+            folder = folder.split("/")[-1]
+            header_dir = Path(folder[0:2])
+            second_lvl_dir = Path(folder[2:4])
+            rest_dir = Path(folder[4:])
 
-        root = Path(input_path / Path(folder))
-        list_of_files = [x for x in root.iterdir() if x.is_file()]
-        print("list of files ", list_of_files)
-
-        try:
-            if not Path.is_dir(output_path / header_dir):
-                Path.mkdir(output_path / header_dir)
-            if not Path.is_dir(output_path / header_dir / second_lvl_dir):
-                Path.mkdir(output_path / header_dir / second_lvl_dir)
-            if not Path.is_dir(output_path / header_dir / second_lvl_dir / rest_dir):
-                Path.mkdir(output_path / header_dir / second_lvl_dir / rest_dir)
+            Path(output_path / header_dir).mkdir(parents=True, exist_ok=True)
+            Path(output_path / header_dir / second_lvl_dir).mkdir(parents=True, exist_ok=True)
+            Path(output_path / header_dir / second_lvl_dir / rest_dir).mkdir(parents=True, exist_ok=True)
 
             Path.rename(input_path / folder, output_path / header_dir / second_lvl_dir / rest_dir)
 
-        except OSError:
-            print("Creation of the directories failed")
-        finally:
-            print("Successfully created the directories")
+    except OSError:
+        print("Creation of the directories failed")
+    finally:
+        print("Successfully created the directories")
 
 
 if __name__ == '__main__':
--- a/scripts/migrate-vfolder-structure.py
+++ b/scripts/migrate-vfolder-structure.py
@@ -6,7 +6,7 @@ def main():
     input_path = Path(sys.argv[1])  # full path for vroot/local
     output_path = Path(sys.argv[2])  # full path for volume directory. ex. vfs/
     print(input_path, output_path)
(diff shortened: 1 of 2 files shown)


scripts/migrate-vfolder-structure.py
```diff
@@ -0,0 +1,41 @@
...
+        list_of_files = os.listdir(input_path + "/" + folder)
+        print(list_of_files)
+        print("list of files ", list_of_files)
+        try:
+            if not os.path.isdir(output_path + "/" + header_dir):
+                os.mkdir(output_path + "/" + header_dir)
+            if not os.path.isdir(output_path + "/" + header_dir + "/" + second_lvl_dir):
+                os.mkdir(output_path + "/" + header_dir + "/" + second_lvl_dir)
+            if not os.path.isdir(output_path + "/" + header_dir + "/" + second_lvl_dir + "/" + rest_dir):
+                os.mkdir(output_path + "/" + header_dir + "/" + second_lvl_dir + "/" + rest_dir)
```
achimnol, line 28: You could do this part with one line: `Path(output_path / header_dir / second_lvl_dir).mkdir(parents=True, exist_ok=True)`

achimnol (changes requested):
Please use [`pathlib`](https://docs.python.org/3/library/pathlib.html) and `/` operator to make it portable.

scripts/migrate-vfolder-structure.py
```diff
@@ -0,0 +1,42 @@
...
+        list_of_files = [x for x in root.iterdir() if x.is_file()]
+        print("list of files ", list_of_files)
+
+        try:
+            if not Path.is_dir(output_path / header_dir):
+                Path.mkdir(output_path / header_dir)
+            if not Path.is_dir(output_path / header_dir / second_lvl_dir):
+                Path.mkdir(output_path / header_dir / second_lvl_dir)
+            if not Path.is_dir(output_path / header_dir / second_lvl_dir / rest_dir):
+                Path.mkdir(output_path / header_dir / second_lvl_dir / rest_dir)
```
achimnol, line 30: Still this part could be reduced to a single line, using `parents=True, exist_ok=True` arguments.

scripts/migrate-vfolder-structure.py
```diff
@@ -0,0 +1,42 @@
...
+from pathlib import Path
+
+
+def main():
+    input_path = Path(sys.argv[1])  # full path for vroot/local
+    output_path = Path(sys.argv[2])  # full path for volume directory. ex. vfs/
+    print(input_path, output_path)
+    subfolders = [x for x in input_path.iterdir() if x.is_dir()]
+
+    print("Subfolders ", subfolders)
```
achimnol, line 11: Should we print all of the vfolder names (they are just a bunch of UUIDs...)?
I think just showing the number of folders would be sufficient information for the user.

scripts/migrate-vfolder-structure.py
```diff
@@ -0,0 +1,42 @@
...
+    for folder in subfolders:
+        folder = str(folder)
+        folder = folder.split("/")[-1]
+        header_dir = Path(folder[0:2])
+        second_lvl_dir = Path(folder[2:4])
+        rest_dir = Path(folder[4:])
+
+        root = Path(input_path / Path(folder))
+        list_of_files = [x for x in root.iterdir() if x.is_file()]
+        print("list of files ", list_of_files)
```
achimnol, line 22: I think this is also unnecessary.
There may be vfolders with many number of files, which would drastically slow down this operation and fill up the user's terminal with those file names.

achimnol (changes requested):
Looks good except the comments.

scripts/migrate-vfolder-structure.py
```diff
@@ -8,33 +8,26 @@ def main():
...
+        for folder in subfolders:
+            folder = str(folder)
+            folder = folder.split("/")[-1]
+            header_dir = Path(folder[0:2])
+            second_lvl_dir = Path(folder[2:4])
+            rest_dir = Path(folder[4:])
+
+            Path(output_path / header_dir).mkdir(parents=True, exist_ok=True)
+            Path(output_path / header_dir / second_lvl_dir).mkdir(parents=True, exist_ok=True)
+            Path(output_path / header_dir / second_lvl_dir / rest_dir).mkdir(parents=True, exist_ok=True)
```
achimnol, line 23: Still this could be reduced to one line, because the parents of the last line's Path includes/implies the first line and the second line.

achimnol (changes requested):

scripts/migrate-vfolder-structure.py
```diff
@@ -0,0 +1,35 @@
...
```
achimnol, line 23: And also... path fragments should be treated as strings, and we can exploit the fact that `output_path` is already a Path object:
```suggestion
            header_dir =folder[0:2]
            second_lvl_dir =folder[2:4]
            rest_dir =folder[4:]

            (output_path / header_dir / second_lvl_dir / rest_dir).mkdir(parents=True, exist_ok=True)
```

scripts/migrate-vfolder-structure.py
leksikov: Completed
(output_path / header_dir / second_lvl_dir / rest_dir).mkdir(parents=True, exist_ok=True)

achimnol (changes requested):
The code looks good now.
Please do the followings for finishing up:
* Add the news fragment as a single-sentence Markdown text file, `changes/319.feature`.
* Please fix up mypy error.
* Update our internal KB1040 to mention this script.

achimnol:
Ok, the mypy error looks like a issue for #321 and follow-up commits. I will merge this PR.

PR #321, opened 2020-09-01
Fix error monitor plugin inheritance

yongkyunlee:
Follow-up of #315 and #316.
- Updated type annotation of context_env in capture_exception


achimnol:
Skipping news fragment as this is a follow-up work of #316.

PR #323, opened 2020-09-04
Support vfolder clone

yongkyunlee:
- Add `cloned_allowed` as a column of `vfolders` table.
- Add endpoints to update vfolder options and clone vfolder.


yongkyunlee:
Dealing with issue lablup/backend.ai#150

PR #329, opened 2020-09-11
Add support for dotfiles of domains and groups

yongkyunlee:
- Add dotfiles column for tables `domains` and `groups`.
- Add manager API endpoints for listing/getting, creating, updating, and deleting dotfiles.

Solving issue lablup/backend.ai#210.


src/ai/backend/gateway/domainconfig.py
```diff
@@ -0,0 +1,194 @@
...
+                resp.append({
+                    'path': entry['path'],
+                    'permission': entry['perm'],
+                    'data': entry['data']
+                })
+            return web.json_response(resp)
+
+
+@server_status_required(READ_ALLOWED)
+@auth_required
```
adrysn, line 112: Domain dotfile should not be updated by normal uses. There should be an authorization check like creation API.

src/ai/backend/gateway/groupconfig.py
```diff
@@ -0,0 +1,252 @@
...
+        t.Key('data'): t.String(max_length=MAXIMUM_DOTFILE_SIZE),
+        t.Key('path'): t.String,
+        t.Key('permission'): t.Regexp(r'^[0-7]{3}$', re.ASCII),
+    }
+))
+async def create(request: web.Request, params: Any) -> web.Response:
+    log.info('CREATE DOTFILE (group: {0})', params['group'])
+    dbpool = request.app['dbpool']
+    if not request['is_superadmin'] and not request['is_admin']:
+        raise GenericForbidden('Only superadmin or admin can create dotfiles')
```
adrysn, line 49: This can be achived by `@admin_required` decorator. Also, group dotfiles should not be created by the admins of a domain where the group does not belong to.

src/ai/backend/gateway/groupconfig.py
```diff
@@ -0,0 +1,252 @@
...
+            if params['domain'] is None:
+                raise InvalidAPIParameters('Missing parameter \'domain\'')
+            query = (sa.select([groups.c.id])
+                        .select_from(groups)
+                        .where(groups.c.domain_name == params['domain'])
+                        .where(groups.c.name == group_id_or_name))
+            group_id = await conn.scalar(query)
+        else:
+            group_id = group_id_or_name
+        if group_id_or_name and group_id is None:
```
adrysn, line 63: `group_id_or_name` does not seem to be needed in this `if` statement.

src/ai/backend/gateway/domainconfig.py
```diff
@@ -0,0 +1,194 @@
...
+            raise DotfileCreationFailed('No leftover space for dotfile storage')
+
+        query = (domains.update()
+                        .values(dotfiles=dotfile_packed)
+                        .where(domains.c.name == params['domain']))
+        await conn.execute(query)
+    return web.json_response({})
+
+
+@auth_required
```
adrysn, line 78: I think it would be better for domain dotfiles not to be queried by domain admins and users in the other domains.

src/ai/backend/gateway/groupconfig.py
```diff
@@ -0,0 +1,252 @@
...
+            raise DotfileCreationFailed('No leftover space for dotfile storage')
+
+        query = (groups.update()
+                        .values(dotfiles=dotfile_packed)
+                        .where(groups.c.id == group_id))
+        await conn.execute(query)
+    return web.json_response({})
+
+
+@auth_required
```
adrysn, line 93: I think it would be better for group dotfiles not to be exposed to users in other groups as well as the admins of domains where the group does not belong to.

src/ai/backend/gateway/groupconfig.py
```diff
@@ -0,0 +1,252 @@
...
+                resp.append({
+                    'path': entry['path'],
+                    'permission': entry['perm'],
+                    'data': entry['data']
+                })
+            return web.json_response(resp)
+
+
+@server_status_required(READ_ALLOWED)
+@auth_required
```
adrysn, line 142: Group dotfiles should only be updatable by superadmins, and the domain admins where the group belongs to.

src/ai/backend/gateway/groupconfig.py
```diff
@@ -0,0 +1,252 @@
...
+            raise DotfileCreationFailed('No leftover space for dotfile storage')
+
+        query = (groups.update()
+                        .values(dotfiles=dotfile_packed)
+                        .where(groups.c.id == group_id))
+        await conn.execute(query)
+    return web.json_response({})
+
+
+@auth_required
```
adrysn, line 191: Any user can delete group dotfiles.

src/ai/backend/manager/registry.py
```diff
@@ -693,12 +693,40 @@ async def enqueue_session(
...
+            dotfile_paths = set(map(lambda x: x['path'], dotfiles))
+            # add keypair dotfiles
+            internal_data.update({'dotfiles': list(dotfiles)})
+            # add group dotfiles
+            query = (sa.select([groups.c.dotfiles])
+                       .select_from(groups)
+                       .where(groups.c.id == group_id))
+            result = await conn.execute(query)
+            row = await result.fetchone()
+            dotfiles = msgpack.unpackb(row['dotfiles'])
```
adrysn, line 711: Seems like the exact same logic with the `query_group_dotfiles`. Why not use it?

src/ai/backend/manager/registry.py
```diff
@@ -693,12 +693,40 @@ async def enqueue_session(
...
+                if dotfile['path'] not in dotfile_paths:
+                    internal_data['dotfiles'].append(dotfile)
+                    dotfile_paths.add(dotfile['path'])
+            # add domain dotfiles
+            query = (sa.select([domains.c.dotfiles])
+                        .select_from(domains)
+                        .where(domains.c.name == domain_name))
+            result = await conn.execute(query)
+            row = await result.fetchone()
+            dotfiles = msgpack.unpackb(row['dotfiles'])
```
adrysn, line 722: How about using `query_domain_dotfiles`?

adrysn (changes requested):
I left some comments mostly on authorizations.

And, I am worried about what will happen when the name of a dotfile collides with the name of a virtual folder. For example, let's say a superadmin created a dotfile named `.test`, which will be mounted at `/home/work/.test` inside a session container. And there is a chance that some users have a virtual folder named `.test`, which will be auto-mounted. The user tries to launch a session. I'm not sure since I didn't test, but probably the session will not be launched due to naming conflict.

This is not very serious for user's dotfiles since the problem will arise only for that user. But, for domain and group dotfiles, the problem will affect many domains' and/or groups' users, if admins are careless with the dotfile naming.

Currently, I don't have a clear idea to prevent this problem. Let's discuss further with others.

adrysn:
LGTM. Please merge the master branch into this one, and resolve conflicting file(s). Also, try to fix broken tests if possible.

adrysn (approved):

PR #333, opened 2020-09-18
Validate service ports before starting session

yongkyunlee:
Although the service ports are validated as they are parsed in the agent, pre-validate them before starting the kernel and raise error appropriately to be passed to the kernel.

Solving issue lablup/backend.ai#220


achimnol (changes requested):
This code is a duplicate from the `ai.backend.agent.utils` module.
Please refactor it out as one of backend.ai-common function using a new module `ai.backend.common.service_ports`.

yongkyunlee:
Used module implemented in PR lablup/backend.ai-common/#52.

Parsing is done twice (once in manager and once in agent), but it is somewhat inevitable. In order to do it once, we would need to modify imageRef and pass parsed service ports to the agent, but this process would be unnecessarily complicated, and there are only few ports to parse in common use cases anyways.

achimnol (approved):

PR #337, opened 2020-10-03
fix: get_task_logs references deprecated etcd config

adrysn:
`get_task_logs` raises an error, so a client cannot get task logs in Backend.AI 20.09. This is due to the handler still uses deprecated etcd volume configs: `volumes/_mount` and `volumes/_fsprefix`. This PR removes them and build `log_path` variable correctly by fetching information from storage-proxy.


achimnol (changes requested):
We need to add a new storage-proxy API to retrieve the file content, assuming that vfolder mounts are NOT available in the manager instances in the future.

PR #346, opened 2020-10-31
fix: Regression of container logs after session ID/name match refactoring

achimnol:


adrysn (changes requested):
Container logs from terminated sessions cannot be fetched. When I see the manager log, it says `No such session`.

image

adrysn (approved):

PR #352, opened 2020-11-12
fix: wrap storage manager request with try-exception block

lizable:
After merging the PR,
uncaught error due to folder related operation failed will be covered by proper exception message.

### DIFF
--- a/src/ai/backend/gateway/vfolder.py
+++ b/src/ai/backend/gateway/vfolder.py
@@ -701,20 +701,23 @@ async def mkdir(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
     folder_name = request.match_info['name']
     access_key = request['keypair']['access_key']
     log.info('VFOLDER.MKDIR (ak:{}, vf:{}, path:{})', access_key, folder_name, params['path'])
-    storage_manager = request.app['storage_manager']
-    proxy_name, volume_name = storage_manager.split_host(row['host'])
-    async with storage_manager.request(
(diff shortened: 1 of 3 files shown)


src/ai/backend/gateway/vfolder.py
```diff
@@ -701,20 +701,23 @@ async def mkdir(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
...
+                'vfid': str(row['id']),
+                'relpath': params['path'],
+                'parents': params['parents'],
+                'exist_ok': params['exist_ok'],
+            },
+            raise_for_status=True,
+        ):
+            pass
+    except aiohttp.ClientResponseError:
+        raise VFolderCreationFailed
```
achimnol, line 720: `VFolderCreationFailed` means the failure of creating a new vfolder, not the failure of creating a directory inside an existing vfolder.
Let's define a new exception `VFolderOperationFailed` in `ai.backend.gateway.exceptions` and use it.
Also, please wrap other storage-manager invocations throughout this module in the same way using proper API-level exception classes.

achimnol (changes requested):

src/ai/backend/gateway/vfolder.py
lizable: I understand.
All of the storage-manager invocation should have exception handling, as you mentioned.

achimnol (approved):
LGTM, but please add the news fragment.

PR #354, opened 2020-11-18
fix: replace config_server legacy code to shared_config

lizable:
This PR is for replacing config_server, which is legacy code to shared_config.


achimnol (approved):
LGTM!

PR #360, opened 2020-12-14
fix: exception on handling broken symlinks in file list API

adrysn:
If there is a symbolic link to a broken target, file list API raises `FileNotFound` exception. This PR changes to not to follow symbolic links, instead just return new `symlink_target` information for possible use case from the client.

This PR should be applied only to 20.03 branch since the API is moved to the storage-proxy in the master branch.


achimnol (approved):
LGTM!

PR #377, opened 2020-12-27
fix: make aiodocker belong to prindipal dependencies

adrysn:
`aiodocker` is used int `registry.py`, but it is put in a test dependency. I think it should be a default dependency.


adrysn:
Already commited in the master branch.

PR #392, opened 2021-01-28
Refactor event bus and related typing

achimnol:


achimnol:
Decided to ignore the test suite failure due to some strange aiopg issues (potentially related to aio-libs/aiopg#647 and aio-libs/aiopg#482), which also look to cause other problems such as inability to support database HA setup.
Since aiopg and asyncpgsa have limited community support recently, let's completely rewrite all database query codes to use the native SQLAlchemy v1.4+ asyncio Core API and eliminate the dependency to aiopg (internal issue: OP#1161).

PR #401, opened 2021-02-24
feat: add BACKENDAI_KERNEL_IMAGE environment variable in container

adrysn:
Some users want to write a script which has different control flow depending on the launched kernel image.


achimnol (approved):
LGTM!

PR #403, opened 2021-03-06
fix: do not register sub containers to existing sessions if main container is TERMINATED

adrysn:
In some cases, sub containers are still RUNNING even though main container is TERMINATED, which we can call broken session. In this case, the scheduler for the scaling group which contains the broken session is unable to schedule PENDING sessions at all. To circumvent this edge case, we exclude broken sessions, that do not have main container, from fetching existing session list.


achimnol (approved):
LGTM!

PR #407, opened 2021-03-24
fix: Add check for pgsql version to adjust connection args

achimnol:
This is a follow-up to #406, to keep the manager compatible with PostgreSQL v10 or older, where the JIT option is not available.


adrysn (approved):
It works in my test environment. I've committed some fixes and check if you are okay with them.

achimnol:
Thanks for extra fixes!

PR #417, opened 2021-04-12
feat: inhibit mounted vfolder deletion

lizable:
After merging this PR, the information about mounted vfolders in each new session follows [MountsExpression](http//github.com/lablup/backend.ai-common/blob/415f94277ee86d2635892aeb732fe78dbee05771/src/ai/backend/common/types.py#L658-L665)


achimnol:
Replaced with lablup/backend.ai#776

PR #422, opened 2021-04-20
fix: Adopt SQLAlchemy-stubs for mypy typechecks with SQLAlchemy codes

achimnol:


achimnol:
This PR is currently "on hold", because sqlalchemy2-stubs still has some issues, such as:
 * It does not allow index-based column access on `Row` objects and requires us to migrate to `row._mapping`. Though, currently it does not yet support PEP646 variadic type arguments (sqlalchemy/sqlalchemy2-stubs#1) to make dot-column-accessors fully type-checked.
 * `AsyncEngine.begin()` method gives confusing `_trans_ctx` instead of `AsyncConnection` (sqlalchemy/sqlalchemy2-stubs#109)

achimnol:
This will be revisited after introduction of async ORMs.

PR #427, opened 2021-04-26
Feature/#424 reflect username

### LINKED ISSUE #424: update operation in username does not reflect.
After applying username update in WebUI and refreshes the page the username goes back as if the change did not happen because the modify user_mutation is for only super admin.

related issue is [here](https://github.com/lablup/backend.ai-webui/issues/957)

Sujin-Kim1:
add update_full_name to update user's full_name regardless of the role (before, the modify user_mutation is for only super admin.)

this closes #424 
it relates to [#957](https://github.com/lablup/backend.ai-webui/issues/957) in webui.

### DIFF
--- a/changes/424.feature
+++ b/changes/424.feature
@@ -1,0 +1,1 @@
+Add update_full_name API to rename user's full_name regardless of role.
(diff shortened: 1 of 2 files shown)


achimnol (approved):
LGTM!

PR #428, opened 2021-05-06
feat:  unify global-default scheduler configs and per-scaling group scheduler options

Sujin-Kim1:
When loading the scheduler, it is to apply both the etcd setting and the current db setting.

this PR related to [#960](https://github.com/lablup/backend.ai-webui/issues/960) in webUI.

### DIFF
--- a/changes/428.feature
+++ b/changes/428.feature
@@ -1,0 +1,1 @@
+- Modify the loading process so that the scheduler can be loaded reflecting scheduler_opts.
--- a/changes/428.feature
+++ b/changes/428.feature
@@ -1,1 +1,1 @@
-- Modify the loading process so that the scheduler can be loaded reflecting scheduler_opts.
+Modify the loading process so that the scheduler can be loaded reflecting scheduler_opts.
--- a/src/ai/backend/manager/scheduler/dispatcher.py
+++ b/src/ai/backend/manager/scheduler/dispatcher.py
@@ -23,6 +23,8 @@ from dateutil.tz import tzutc
 import sqlalchemy as sa
 from sqlalchemy.ext.asyncio import AsyncConnection as SAConnection
 from sqlalchemy.sql.expression import true
+from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.orm.exc import MultipleResultsFound
 
 from ai.backend.common.events import (
     AgentStartedEvent,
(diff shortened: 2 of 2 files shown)


changes/428.feature
```diff
@@ -0,0 +1 @@
+- Modify the loading process so that the scheduler can be loaded reflecting scheduler_opts.
```
achimnol, line 1: This news fragment string is merged into a Markdwon-formatted list when released, via [towncrier](https://pypi.org/project/towncrier/). So you don't have to put the starting `- ` because `* ` will be added automatically. :wink:

achimnol, line 1: You don't have to put the starting `- ` since `* ` will be automatically added by [towncrier](https://pypi.org/project/towncrier/) when generating the changelog. Please remove it. :wink:

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -23,6 +23,8 @@
 import sqlalchemy as sa
 from sqlalchemy.ext.asyncio import AsyncConnection as SAConnection
 from sqlalchemy.sql.expression import true
+from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.orm.exc import MultipleResultsFound
```
achimnol, line 27: Merge into one line because the import source is same.
```suggestion
from sqlalchemy.orm.exc import MultipleResultsFound, NoResultsFound
```

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -242,14 +244,22 @@ async def _load_scheduler(
...
+        try:
+            query = (
+                sa.select([scaling_groups.c.scheduler, scaling_groups.c.scheduler_opts])
+                .select_from(scaling_groups)
+                .where(scaling_groups.c.name == sgroup_name)
+            )
+            result = await db_conn.execute(query)
+            row = result.one()
+            scheduler_name = row['scheduler']
+            scheduler_opts = row['scheduler_opts']
```
achimnol, line 256: To distinguish with the final merge result, this variable may be named differently, e.g., `local_scheduler_opts`.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -242,14 +244,22 @@ async def _load_scheduler(
...
+        except MultipleResultsFound as e:
+            log.exception('_load_scheduler(): multiple scheduler! \n{}', repr(e))
```
achimnol, line 258: When logging exceptions inside the `except` block, you don't have to manually serialize the exception.
```suggestion
            log.exception("_load_scheduler(): multiple scheduler!")
```
will do the what you have intendend.
In Python's `except` blocks, the runtime automatically sets `sys.exc_info()` tuple and most exception-handling utilities including loggers are able to recognize it.
And although this is not a strict rule yet, please use double-quotes for user-facing strings. In the future we may migrate to use the black style and it's worth to try minimizing diffs for new codes.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -242,14 +244,22 @@ async def _load_scheduler(
...
+        except NoResultFound as e:
+            log.exception('_load_scheduler(): scheduler does not exist! \n{}', repr(e))
```
achimnol, line 260: Same here!

achimnol, line 260: It is a good practice to handle potential errors explicitly, but in this case you may assume that there will be always a single row, because the `name` column in the `scaling_groups` table is the primary key and this function is already called from an iteration loop over the list of existing ("schedulable") scaling groups.
(FYI: "schedulable" here means that we consider only the scaling groups which has at least one ALIVE agent)

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -242,14 +244,22 @@ async def _load_scheduler(
...
+        scheduler_opts = {**self.shared_config['plugins']['scheduler'][scheduler_name], **scheduler_opts}
```
achimnol, line 261: What happens if the etcd does not include configs for the given scheduler at all? Does this handle that case as an empty dict?

achimnol (changes requested):

achimnol:
I'd also suggest to elaborate the PR title a little more since "session creation retries" looks like it's about adding a new feature to retry session creation requests when there are session creation failures, though it's actually about implementing a unified view of the global-default scheduler configs and per-scaling group scheduler options. The config/options may include session creation retry thresholds internally, but this PR is *not* about itself.

achimnol (approved):
LGTM!

PR #432, opened 2021-05-20
Feature/utilization based idle session checker

leksikov:
Using idle checker framework, implement resource collecting feature by terminating session based on following things:
- utilization
- Metric: CPU, GPU, RAM...
- Default rule: To be determined

The current implementation uses default thresholds from idle config file for CPU and Memory utilization.


src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,133 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+        {
+            t.Key("threshold", default="10m"): tx.TimeDuration(),
+        }
+    ).allow_extra("*")
+
+    _evhandlers: List[EventHandler[None, AbstractEvent]]
+
+    async def populate_config(self, raw_config: Mapping[str, Any]) -> None:
+        config = self._config_iv.check(raw_config)
+        self.idle_timeout = config["threshold"]
```
adrysn, line 323: Do we use `checkers/timeout/threshold` for the utilization check period? If this is needed, wouldn't it be better to separate out as `checkers/utilization/check-period` or something like that? In case there is an admin who wants to enable utilization checker, but not idle timeout, still we need to write both of the configs in the current implementation.

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,133 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+            f"session.{session_id}.last_access", "0", exist=self._redis.SET_IF_EXIST
+        )
+
+    async def _update_timeout(self, session_id: SessionId) -> None:
+        log.debug(f"ExecutionIdleChecker._update_timeout({session_id})")
+        t = await self._redis.time()
+        await self._redis.set(
+            f"session.{session_id}.last_access",
+            f"{t:.06f}",
+            expire=max(86400, int(self.idle_timeout.total_seconds() * 2)),
```
adrysn, line 356: `idle_timeout` does not seem to be a valid property name since the utilization checker does not garbage collect compute session by the `idle_timeout`. Here, it is just a check period for the utilization.

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,133 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+            self.cpu_util_series.append(float(cpu_util_pct))
+            self.mem_util_series.append(float(mem_util_pct))
+            print(self.cpu_util_series, len(self.cpu_util_series))
+            return True
+        else:
+            avg_cpu_util = sum(self.cpu_util_series) / len(self.cpu_util_series)
+            avg_mem_util = sum(self.mem_util_series) / len(self.mem_util_series)
+
+            self.cpu_util_series = []
+            self.mem_util_series = []
```
adrysn, line 429: Just clear the collected time-series utilization data when they overflow the window size? Wouldn't it be natural to pop out the oldest element from the data and append the latest one?

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,133 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+
+            if (avg_cpu_util <= self.cpu_thresh) or (avg_mem_util <= self.mem_thresh):
```
adrysn, line 431: There may be a case when an admin wants to garbage collect sessions when both CPU **AND** memory utilization lie below the threshold. Can we introduce a config item for this scenario?

adrysn (changes requested):
I just read through the codes and leave my first impressions. If you have your reason for each item, please feel free to express your opinion. I'll leave other comments after I actually test whether the utilization-based timeout works.

src/ai/backend/manager/idle.py
adrysn: Also, the raw_config here does not carry the `threshold` item since `raw_config` should be a config mapping under `/config/idle/utilization`.

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,133 @@ class UtilizationIdleChecker(BaseIdleChecker):
     """
 
     name: ClassVar[str] = "utilization"
-    default_cpu_util_threshold: ClassVar[float] = 30.0
-    default_accelerator_util_threshold: ClassVar[float] = 10.0
-    default_initial_grace_period: ClassVar[float] = 300.0  # allow first 5 minutes to be idle
```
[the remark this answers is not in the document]
leksikov: grace-period is for delaying session-checker?
I think we may not need the grace-period decision of idle execution based on multiple data points in a window so that early values should not have much impact on average overall. Also, early points get popped out first as time passes.

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,133 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
 
-    async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
-        # last_stat = session['last_stat']
-        # TODO: implement
-        return True
+    _config_iv = t.Dict(
+        {
+            t.Key("threshold", default="10m"): tx.TimeDuration(),
+        }
+    ).allow_extra("*")
```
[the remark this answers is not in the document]
adrysn: I think this `_config_iv` should check all the essential Etcd items under `config/idle/checkers/utilization`.

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,148 @@ class UtilizationIdleChecker(BaseIdleChecker):
     """
 
     name: ClassVar[str] = "utilization"
-    default_cpu_util_threshold: ClassVar[float] = 30.0
-    default_accelerator_util_threshold: ClassVar[float] = 10.0
-    default_initial_grace_period: ClassVar[float] = 300.0  # allow first 5 minutes to be idle
+    cpu_util_series: List[float] = []
+    mem_util_series: List[float] = []
+    cuda_mem_util_series: List[float] = []
```
adrysn, line 310: I think CUDA core utilization should also be taken into account.

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,148 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+            f"session.{session_id}.last_execution", "0", exist=self._redis.SET_IF_EXIST
+        )
+
+    async def _update_timeout(self, session_id: SessionId) -> None:
+        log.debug(f"ExecutionIdleChecker._update_timeout({session_id})")
+        t = await self._redis.time()
+        await self._redis.set(
+            f"session.{session_id}.last_execution",
+            f"{t:.06f}",
+            expire=float(sum(self.cpu_util_series) / len(self.cpu_util_series))
```
adrysn, line 349: I didn't understand this line. Why the expiration of redis key is dependent on CPU utilization history?

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,148 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+        interval = self.timer.interval
+
+        if "m" in self.window:
+            window = int(self.window.replace("m", ""))
+            window_size = (window * 60) / interval
+        elif "h" in self.window:
+            window = int(self.window.replace("h", ""))
+            window_size = (window * 60 * 60) / interval
+        else:
+            raise Exception("Undefined unit of window interval")
```
adrysn, line 416: Please refer to the utilization function to convert humanized string to `datetime` object.
https://github.com/lablup/backend.ai-common/blob/main/src/ai/backend/common/utils.py#L144

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,148 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+
+        if len(self.cpu_util_series) < window_size:
+            self.cpu_util_series.append(float(cpu_util_pct))
+            self.mem_util_series.append(float(mem_util_pct))
+            self.cuda_mem_util_series.append(float(cuda_mem_util_pct))
+            return True
+        else:
```
adrysn, line 423: Don't we need to append `*_util_pct` to `self.*_util_series` in `else` block as well? I can see `pop`ing oldest elements out, but not appending new values.

adrysn (changes requested):
I left some comments. Also, please fix listing errors.

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +305,148 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+    _evhandlers: List[EventHandler[None, AbstractEvent]]
 
-    async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
-        # last_stat = session['last_stat']
-        # TODO: implement
-        return True
+    async def populate_config(self, raw_config: Mapping[str, Any]) -> None:
+        self.cpu_threshold = int(raw_config["resource-thresholds"]["cpu"]["average"])
+        self.mem_threshold = int(raw_config["resource-thresholds"]["mem"]["average"])
+        self.cuda_mem_threshold = int(raw_config["resource-thresholds"]["cuda_mem"]["average"])
```
adrysn, line 317: There is "OR" check condition, but Manager fails to be launched when not all of `cpu`, `mem`, and `cuda_mem` are configured. For example, I would like to garbage collect compute sessions by CPU utilization criterion only, but I was unable to launch Manager with the following `config/idle` setting:

```json
{
    "app-streaming-packet-timeout": "5m",
    "checkers": {
        "timeout": {
            "threshold": "24h"
        },
        "utilization": {
            "resource-thresholds": {
                "cpu": {
                    "average": "10",
                    "window": "30s"
                }
            },
            "thresholds_check_condition": {
                "condition": "and"
            }
        }
    },
    "enabled": "timeout,utilization"
}
```

src/ai/backend/manager/config.py
```diff
@@ -80,7 +80,12 @@
                - window: "1m"  # average for duration
                # NOTE: To use "cuda.mem" criteria, user programs must use
                #       an incremental allocation strategy for CUDA memory.
```
adrysn, line 82: In example configs, please add `"mem"` threshold as well.

adrysn (changes requested):
Some comments on resource configs.

src/ai/backend/manager/config.py
```diff
@@ -80,7 +80,12 @@
...
-           - initial-grace_period: "30s"
+           + thresholds_check_condition:
+               - condition: "and" # logical operator
```
adrysn, line 84: Will there be more sub-items under `thresholds_check_condition`? If not, how about to set the condition under `thresholds-check-condition` instead of `thresholds_check_condition/condition`? Also, please use dash (`-`), not underscore (`_`) for consistency with other key names (such as `resource-thresholds` and `app-streaming-packet-timeout`.

adrysn (commented):
Comments on config naming.

src/ai/backend/manager/idle.py
leksikov: Overall, here is implemented the sliding moving window with step 1. After window fills up until defined window size. It will be always popping the oldest value and adding the newest value. So that window moves by 1 element.  And in that window we continuously measure the average and if the threshold is reached. There is no loop, but session_checker()  repetitively called by _do_idle_checker(), so the adding and removing elements implemented in if/else block.

The execution of the code is following:
If a number of elements are less than the defined window size then append a new element from Redis and return True.
Since returned value True doesn't terminate the session, the session_checker() is called again by _do_idle_check().
It checks the window size and sees that it again has fewer elements than the defined window size. So it appends and session_checker is called until a number of elements become equal to the defined window size.

At that point, in Else condition. The average is measured for all values in a window. Now we don't need the oldest element anymore, so it gets popped out. Then we check if the threshold condition is met. If a threshold is not met return True value. After that session_checker will be triggered again by _do_idle_check() and since a number of elements in the window are less than the defined window size, then the newest value will be appended and then Else condition will be triggered, and the threshold will be again checked.

src/ai/backend/manager/idle.py
leksikov: I had just a draft idea of using CPU utilization as a proxy for time measure, to make the expiration record more dynamic according to resource usage. The more CPU utilized the more time is given for expiration record. However, I didn't come out with a solution of scale range definition that maps cpu usage to time. Now, I consider to set expiration to 86400 seconds as the default value for storing record.

src/ai/backend/manager/idle.py
leksikov: Yes, I can solve this issue. I will write a lot of if/else codes to cover all kinds of configurations, unless I find a better alternative for multiple if/else code statements.

src/ai/backend/manager/idle.py
adrysn: Yes, I understand what you mean, and the logic will work. But, I wonder is there a reason not to append new data in the `else` block and not going back to the `if` block any more once the window size is reached.

Also, the `else` keyword does not seem to be needed here since the `if` block always returns `True`. So, what do you think of the slightly changed logic below?

```python
self.cpu_util_series.append(float(cpu_util_pct))
self.mem_util_series.append(float(mem_util_pct))
self.cuda_mem_util_series.append(float(cuda_mem_util_pct))
if len(self.cpu_util_series) < window_size:
    return True

self.cpu_util_series.pop(0)
self.mem_util_series.pop(0)
avg_cpu_util = sum(self.cpu_util_series) / len(self.cpu_util_series)
avg_mem_util = sum(self.mem_util_series) / len(self.mem_util_series)
cuda_mem_util = sum(self.cuda_mem_util_series) / len(self.cuda_mem_util_series)
if self.threshold_condition == "or":
    ...
```

src/ai/backend/manager/idle.py
adrysn: Aha. I think your idea is great. The only problem here is that there should be a situation that CPU utilization is turned off. Some admins may want to garbage collect session based only on the GPU utilization. In that case, the CPU utilization has no meaning in terms of dynamic expiration time.

If you can introduce some logic of generalized utilization or something else, I agree with your idea of dynamic expiration time. Or, you can make it configurable by introducing a Etcd key.

src/ai/backend/manager/idle.py
adrysn: I think you can take out the utilization calculation logic to a new internal method. Then, you may still write simpler code even if there are many independent resources to be considered. For a no-brainer mock example:

```python
cpu_util_pct = float(live_stat["cpu_util"]["pct"])
mem_util_pct = float(live_stat["mem"]["pct"])
cuda_mem_util_pct = float(live_stat["cuda_mem"]["pct"])

cpu_check_result = _check_resource('cpu', cpu_util_pct)
mem_check_result = _check_resource('mem', mem_util_pct)
cuda_mem_check_result = _check_resource('cuda_mem', cuda_mem_util_pct)
# If async function is needed, you may create tasks and use asyncio.gather to collect every results

if self.threshold_condition == "or":
  if cpu_check_result or mem_check_result or cuda_mem_check_result:
    return ...
...

def _check_resource(util_series_name, util_pct):
  interval = self.timer.interval
  window_size = ...
  util_series = getatr(self, f'{util_series_name}')
  if len(util_series) < window_size:
    setattr(self, f'{util_series_name}_util_series', util_series.append(float(cpu_util_pct)))
    return True
  avg_util = ...
```

adrysn: Also, for easier fetching of values from deeply-nested dict, you can use `nmget` function in `common` module.
https://github.com/lablup/backend.ai-common/blob/main/src/ai/backend/common/utils.py#L105

```python
from ai.backend.common.utils import nmget
self.cpu_threshold = int(nmget(raw_config, "resource-thresholds.cpu.average"))
self.mem_threshold = int(nmget(raw_config, "resource-thresholds.mem.average"))
self.cuda_mem_threshold = int(nmget(raw_config, "resource-thresholds.cuda_mem.average"))
```

src/ai/backend/manager/idle.py
```diff
@@ -307,24 +318,196 @@ class UtilizationIdleChecker(BaseIdleChecker):
...
+
+    async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
+        session_id = session["id"]
+        active_streams = await self._redis.zcount(
+            f"session.{session_id}.active_app_connections"
+        )
+        if active_streams is not None and active_streams > 0:
+            return True
+
+        redis = await aioredis.create_redis("redis://127.0.0.1:8111/")
```
adrysn, line 436: I am having a hard time understanding this code. Obviously, the hard-coded address should be changed to a config item, but I am curious why is it needed to create a Redis connection when `BaseIdleChecker` has already created one in `__ainit__`?

leksikov: Redis connection based on address in config item doesn't return any values. The url for the config redis address is _redis://127.0.0.1:8111/2_
I needed to work with actual data to move forward, so made redis connection with the appropriate url address.

adrysn: In that case, you should use [DB number 0 (STAT_DB)](https://github.com/lablup/backend.ai-manager/blob/main/src/ai/backend/manager/defs.py#L27), so the hard-coded string should be replaced something like:

```python
from ai.backend.manager.defs import REDIS_STAT_DB
...
await aioredis.create_redis(
    str(self._shared_config.get_redis_url(db=REDIS_STAT_DB))
)
```

src/ai/backend/manager/idle.py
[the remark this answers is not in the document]
leksikov: I don't have knowledge about what kind of type is suitable for the execution variables. Is it custom defined type or float?
Also, I use None type for checking available resources to construct boolean conditions since etcd returns None if there is no specified resource.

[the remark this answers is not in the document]
adrysn: I added `_config_iv` for utilization idle checker class.

adrysn:
I think this PR can be improved by saving utilization histories, such as `self.cpu_util_series`, in Redis instead of keeping them as properties of the `UtilizationIdleChekcer` class. When Manager restarts, those properties are initialized and we have to wait for the time windows to be filled again. Note that Manager restarts are much more frequent than Redis restarts.

I'll try to refactor the saving logic a little during the weekends.

adrysn:
> I think this PR can be improved by saving utilization histories, such as `self.cpu_util_series`, in Redis instead of keeping them as properties of the `UtilizationIdleChekcer` class. When Manager restarts, those properties are initialized and we have to wait for the time windows to be filled again. Note that Manager restarts are much more frequent than Redis restarts.
> 
> I'll try to refactor the saving logic a little during the weekends.

Actually, utilization histories SHOULD be saved to Redis since the current implementation will not work on high-availability manager configuration.

src/ai/backend/manager/idle.py
```diff
@@ -326,6 +378,147 @@ async def update_app_streaming_status(
...
+        current_utilizations = await self.get_current_utilization(kernel_ids, occupied_slots)
+        if current_utilizations is None:
+            return True
+
+        # Update utilization time-series data.
+        not_enough_data = False
+        util_series_key = f"session.{session_id}.util_series"
+        raw_util_series = await self._redis.get(util_series_key)
+        try:
+            util_series = json.loads(raw_util_series)
```
achimnol, line 450: I'd suggest to use `msgpack` for faster serialization & deserialization and less memory usage in Redis, unless there are specific reasons to make it human-readable in the memory.

src/ai/backend/manager/idle.py
```diff
@@ -302,22 +318,58 @@ async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
...
-    Checks the idleness of a session by the current utilization of all
-    compute devices and agents assigned to it.
+    Checks the idleness of a session by the average utilization of compute devices.
     """
 
     name: ClassVar[str] = "utilization"
-    default_cpu_util_threshold: ClassVar[float] = 30.0
-    default_accelerator_util_threshold: ClassVar[float] = 10.0
-    default_initial_grace_period: ClassVar[float] = 300.0  # allow first 5 minutes to be idle
+    check_period: float = 20.0
```
achimnol, line 325: `check_interval` would be better than `check_period`.
Period usually means the time length of an action, while interval usually means the time length between two actions.
(e.g., I visit here for 1 week every 1 month. -> period = 1 week, interval = 1 month)

src/ai/backend/manager/idle.py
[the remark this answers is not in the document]
achimnol: Hm.. I think it's better to keep the initial grace period, if we want to **apply the utilization idle checker for *interactive* sessions**. Otherwise, it may be removed.

src/ai/backend/manager/idle.py
```diff
@@ -66,6 +67,7 @@ class AppStreamingStatus(enum.Enum):
 class BaseIdleChecker(aobject, metaclass=ABCMeta):
 
     name: ClassVar[str] = "base"
+    check_period: float = 10.0
 
     _db: SAEngine
```
achimnol, line 72: Let's update the type annotation here: `ai.backend.manager.models.utils.ExtendedAsyncSAEngine`.

src/ai/backend/manager/idle.py
```diff
@@ -126,22 +132,25 @@ async def _do_idle_check(
         source: AgentId,
         event: DoIdleCheckEvent,
     ) -> None:
-        log.debug('do_idle_check(): triggered')
+        log.debug("do_idle_check({}, {}s): triggered", self.name, str(self.check_period))
         async with self._db.begin() as conn:
```
achimnol, line 136: Let's replace `begin()` with `begin_readonly()` because idle checkers will only read the database.

achimnol (changes requested):

src/ai/backend/manager/idle.py
leksikov: msgpack can't unpack this type of stored data with values in the list
{'cpu': [0.42, 0.44], 'cuda_mem': [0.0, 0.0], 'cuda_util': [0.0, 0.0], 'mem': [30.0, 30.0]}

achimnol: Fixed the inability to unpack list values as lists in lablup/backend.ai-common#79 . Now you may pass `use_list=True` to our msgpack wrapper functions to preserve lists for history updates.

src/ai/backend/manager/idle.py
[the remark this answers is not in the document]
leksikov: Added grace period. check_session will continue execution after grace period time.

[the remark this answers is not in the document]
adrysn: Since the utilization checker does not terminate a session until the length of the collected data exceeds `window-size`. So, I'm not sure `initial-grace-period` is required. But, there may be use cases to set longer `initial-grace-period` with shorter `window-size`, so I don't have an objection to the introduction of the `initial-grace-period`. So, I'll mark this issue as resolved.

adrysn:
I found one (possible) problem. When GPU utilization is configured, a no-GPU session is always garbage collected by GPU utilization criterion. I think we need to treat non-GPU sessions as an exception. What do you think? @leksikov @achimnol

achimnol:
> I found one (possible) problem. When GPU utilization is configured, a no-GPU session is always garbage collected by GPU utilization criterion. I think we need to treat non-GPU sessions as an exception. What do you think? @leksikov @achimnol

We need to consider only resource slots that are allocated non-zero.

adrysn:
> > I found one (possible) problem. When GPU utilization is configured, a no-GPU session is always garbage collected by GPU utilization criterion. I think we need to treat non-GPU sessions as an exception. What do you think? @leksikov @achimnol
> 
> We need to consider only resource slots that are allocated non-zero.
> This means that the history fields in Redis should be somehow mapped with resource slot names.

I agree. Can you implement not to garbage collect sessions without GPU even when GPU utilization idle checker is running? @leksikov

leksikov:
> > > I found one (possible) problem. When GPU utilization is configured, a no-GPU session is always garbage collected by GPU utilization criterion. I think we need to treat non-GPU sessions as an exception. What do you think? @leksikov @achimnol
> > 
> > 
> > We need to consider only resource slots that are allocated non-zero.
> > This means that the history fields in Redis should be somehow mapped with resource slot names.
> 
> I agree. Can you implement not to garbage collect sessions without GPU even when GPU utilization idle checker is running? @leksikov

Yes. I can implement this by getting available resources from the session. And then ignore utilization for absent resources

src/ai/backend/manager/idle.py
```diff
@@ -326,6 +384,164 @@ async def update_app_streaming_status(
...
+        try:
+            utilizations = {k: 0.0 for k in self.resource_thresholds}
+            for kernel_id in kernel_ids:
+                raw_live_stat = await self._redis_stat.get(str(kernel_id), encoding=None)
+                live_stat = msgpack.unpackb(raw_live_stat)
+                kernel_utils = {
+                    k: float(nmget(live_stat, f"{k}.pct", 0.0))
+                    for k in self.resource_thresholds
+                }
+                kernel_utils["cpu"] = float(nmget(live_stat, "cpu_util.pct", 0.0))
```
adrysn, line 524: I there any reason to change previous `resource-thresholds.cpu_util` Etcd key to `resource-thresholds.cpu`, and adding this line to explicitly convert `cpu_util` in Redis to `cpu`?

src/ai/backend/manager/idle.py
```diff
@@ -302,22 +321,61 @@ async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
...
     name: ClassVar[str] = "utilization"
-    default_cpu_util_threshold: ClassVar[float] = 30.0
-    default_accelerator_util_threshold: ClassVar[float] = 10.0
-    default_initial_grace_period: ClassVar[float] = 300.0  # allow first 5 minutes to be idle
+    check_interval: float = 20.0
+
+    _config_iv = t.Dict(
+        {
+            t.Key("time-window", default="10m"): tx.TimeDuration(),
+            t.Key("initial-grace-period", default="5m"): tx.TimeDuration(),
```
adrysn, line 333: You introduced `initial-grace-period` configs in Etcd, but I cannot find any code to reference this config.

src/ai/backend/manager/idle.py
```diff
@@ -302,22 +321,61 @@ async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
...
-        # TODO: implement
-        return True
+    async def aclose(self) -> None:
+        self._redis_stat.close()
+        await self._redis_stat.wait_closed()
+        await super().aclose()
+
+    async def populate_config(self, raw_config: Mapping[str, Any]) -> None:
+        config = self._config_iv.check(raw_config)
+        config
```
adrysn, line 366: Delete this line.

adrysn (changes requested):
Comments on recent commits.

src/ai/backend/manager/idle.py
leksikov: I implemented grace period, by that time the issue was resolved without needing the grace period. So that, I was deleting my grace period implementation. Currently, deleted all lines of grace period.

src/ai/backend/manager/idle.py
leksikov: At that time I was trying to check the execution of idle checker and found bugs while trying to execute idle checker.

According to the history of your latest commit, I found that in config.py there is a _cpu_ key value setting. So, trafaret was giving an error that _cpu is not valid key_. So, I changed in resource-thresholds.cpu_util to resource-thresholds.cpu to match with etcd config.py settings.

For adding explicitly, the Redis live_stat output doesn't contain the cpu key. It contains cpu_util, so that _k: float(nmget(live_stat, f"{k}.pct", 0.0))_ line of code wasn't able to access cpu values under the key _cpu_util_ and cpu values in stream were zero. 

Maybe settings all keys to cpu_util including the config.py to match Redis data stream is better? Then there will be not needed for explicit cpu assignment line.

adrysn: Yes. I think it would be better to change the Etcd key to `cpu_util` than adding a code to handle an exception.

src/ai/backend/manager/idle.py
adrysn: I might get lost, but didn't we determined to introduce the `initial-grace-period` to support use cases for a longer initial-grace-period with a shorter time-window?

and, `config` in line 366 above should be deleted.

leksikov: Oh, I have misunderstood your comment before about introducing the grace period.
 I will add grace period configs and implementation soon.
 
 """Since the utilization checker does not terminate a session until the length of the collected data exceeds window-size. So, I'm not sure initial-grace-period is required. But, there may be use cases to set longer initial-grace-period with shorter window-size, so I don't have an objection to the introduction of the initial-grace-period. So, I'll mark this issue as resolved."""

leksikov: Reintroduced initial grace period

src/ai/backend/manager/idle.py
leksikov: Completed making cpu_util consistent

src/ai/backend/manager/idle.py
```diff
@@ -368,6 +371,10 @@ async def populate_config(self, raw_config: Mapping[str, Any]) -> None:
         }
         self.thresholds_check_operator = config.get("thresholds-check-operator")
         self.time_window = config.get("time-window")
+        self.initial_grace_period = config.get("initial-grace-period")
+        self.time_pass_grace_period = self.initial_grace_period.total_seconds() \
+            + (await self._redis.time())
+        self.grace_period_flag = False
```
adrysn, line 377: Hmm... The grace period should be applied session-wise. Each session should not be terminated at least for the duration of the grace period. Since `populate_config` is called only once at the start of the Manager, every session is subject to be garbage collected after the `initial-grace-period` passed upon Manager restart. To get the per session grace timeout, you may save the information on Redis or just reference the session's `created_at` information.

Also, `self.grace_period_flag` has no role in the current code.

adrysn (changes requested):
The current implementation of the `initial-grace-period` does not apply to each session. It is a rather per-Manager setting.

src/ai/backend/manager/idle.py
adrysn: The problem is resolved by lablup/backend.ai-common#79, but it was not backported to 21.03. So, I backported by
https://github.com/lablup/backend.ai-common/commit/ed4d567f841340f3612e34a215c007d19311f77f, and I think we can use msgpack to pack/unpack the utilization series data.

adrysn (approved):

adrysn:
I pushed some commits today and I think this PR now works as I expected.

@leksikov Please check one last time for every feature of this PR. Especially, check if `cuda_util` criterion working fine. I checked `cpu_util` works, but not `cuda_util` since `cuda_util` and `cuda_mem` are always zero in CentOS 7.5, which is the OS of my test node. If you think all is okay, we'll merge this PR.

@achimnol After @leksikov confirms the PR, please merge it unless you have other issues to raise.

leksikov:
Fixed cuda_util and cuda_mem presence in unavailable resources. Since the beginning 'cuda.device' and 'cuda.shares' are both presents in the occupied_slots dictionary. However, only one of them is active for the session. However, since one of them is not active the cuda_util and cuda_mem gets into the unavaileble_resourses list resulting the GPU condition never gets checked. My solution checks for presence for Cuda device in occupied slots and then check if cuda.device or cuda.shares is active and corrects the unavailable_resourse list accordingly.

adrysn:
> Fixed cuda_util and cuda_mem presence in unavailable resources. Since the beginning 'cuda.device' and 'cuda.shares' are both presents in the occupied_slots dictionary. However, only one of them is active for the session. However, since one of them is not active the cuda_util and cuda_mem gets into the unavaileble_resourses list resulting the GPU condition never gets checked. My solution checks for presence for Cuda device in occupied slots and then check if cuda.device or cuda.shares is active and corrects the unavailable_resourse list accordingly.

Good point!

src/ai/backend/manager/idle.py
```diff
@@ -422,6 +422,11 @@ async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
             if occupied_slots[slot] == 0:
                 unavailable_resources.update(self.slot_resource_map[slot])
 
+        if ("cuda.device" in occupied_slots.keys()) or ("cuda.shares" in occupied_slots.keys()):
+            if (occupied_slots["cuda.device"] == 1) or (occupied_slots["cuda.shares"] == 1):
```
adrysn, line 426: This line will raise an error if `occupied_slots` have only one of `cuda.device` or `cuda.shares`.

adrysn (changes requested):

src/ai/backend/manager/idle.py
adrysn: If `occupied_slots["cuda.shares"]` is not zero and `1`, what will happen?

adrysn: We also need to be as much as general here since there can be many other types of CUDA slots, for example, "cuda.mig.*", which may be present in MIG-enabled environment.

adrysn: I made a commit to resolve the above-mentioned issues. Please check if this works.
https://github.com/lablup/backend.ai-manager/pull/432/commits/14266d77231e1dde148511f19b50219cb3e25122

src/ai/backend/manager/idle.py
```diff
@@ -326,6 +394,171 @@ async def update_app_streaming_status(
...
+            util_series_key,
+            msgpack.packb(util_series),
+            expire=max(86400, int(self.time_window.total_seconds() * 2)),
+        )
+        if not_enough_data:
+            return True
+
+        # Check over-utilized (not to be collected) resources.
+        avg_utils = {k: sum(util_series[k]) / len(util_series[k]) for k in util_series}
+        over_utilized = {
```
achimnol, line 507: How about renaming this to `sufficiently_utilized`?

src/ai/backend/manager/idle.py
```diff
@@ -326,6 +394,171 @@ async def update_app_streaming_status(
...
+
+        # Update utilization time-series data.
+        not_enough_data = False
+        util_series_key = f"session.{session_id}.util_series"
+        raw_util_series = await self._redis.get(util_series_key, encoding=None)
+
+        try:
+            util_series = msgpack.unpackb(raw_util_series, use_list=True)
+        except TypeError:
+            util_series = {k: [] for k in self.resource_thresholds}
```
achimnol, line 489: To highlight that `self.resource_thresholds` is a dict, let's use `.keys()` here and in similar occurrences afterwards.

src/ai/backend/manager/idle.py
```diff
@@ -326,6 +394,171 @@ async def update_app_streaming_status(
...
+        )
+        if not_enough_data:
+            return True
+
+        # Check over-utilized (not to be collected) resources.
+        avg_utils = {k: sum(util_series[k]) / len(util_series[k]) for k in util_series}
+        over_utilized = {
+            k: (float(avg_utils[k]) >= float(self.resource_thresholds[k]))
+            for k in self.resource_thresholds
+            if (self.resource_thresholds[k] is not None) and (k not in unavailable_resources)
```
achimnol, line 510: We could use `for k, threshold in self.resource_thresholds.items()` for brevity and to avoid duplicate lookups.

src/ai/backend/manager/idle.py
```diff
@@ -302,22 +325,67 @@ async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
...
-    default_cpu_util_threshold: ClassVar[float] = 30.0
-    default_accelerator_util_threshold: ClassVar[float] = 10.0
-    default_initial_grace_period: ClassVar[float] = 300.0  # allow first 5 minutes to be idle
+    check_interval: float = 20.0
+
+    _config_iv = t.Dict(
+        {
+            t.Key("time-window", default="10m"): tx.TimeDuration(),
+            t.Key("initial-grace-period", default="5m"): tx.TimeDuration(),
+            t.Key("thresholds-check-operator", default="and"): t.String,
```
achimnol, line 338: Let's use `tx.Enum` so that the validation happens earlier, by defining an enum class to represent the available operation types.
(ref: https://github.com/lablup/backend.ai-common/blob/main/src/ai/backend/common/validators.py#L171)

src/ai/backend/manager/idle.py
```diff
@@ -326,6 +394,171 @@ async def update_app_streaming_status(
...
+            # NOTE: Manual calculation of mem utilization.
+            # mem.capacity does not report total amount of memory allocated to
+            # the container, and mem.pct always report >90% even when nothing is
+            # executing. So, we just replace it with the value of occupied slot.
+            mem_slots = float(occupied_slots.get('mem', 0))
+            mem_current = float(nmget(live_stat, "mem.current", 0.0))
+            utilizations['mem'] = mem_current / mem_slots * 100 if mem_slots > 0 else 0
+            return utilizations
+        except Exception as e:
+            log.warning("Unable to collect utilization for idleness check:{}", repr(e))
```
achimnol, line 559:
```suggestion
            log.warning("Unable to collect utilization for idleness check", exc_info=e)
```

achimnol (changes requested):
Just added a few code-level comments. :wink:

src/ai/backend/manager/idle.py
leksikov: renamed to sufficiently_utilized

src/ai/backend/manager/idle.py
leksikov: using keys for self.resource_thresholds

src/ai/backend/manager/idle.py
leksikov: changed to _for k, threshold in self.resource_thresholds.items()_

src/ai/backend/manager/idle.py
leksikov: changed the warning printout

src/ai/backend/manager/idle.py
leksikov: added t.Enum('and', 'or') for thresholds-check-operator

leksikov: Changed Operator class to ThresholdOperator

achimnol (approved):
LGTM!

PR #437, opened 2021-06-11
feat: add access key into environment

lizable:
This PR adds each user's own access key as an environment variable called 'BACKENDAI_ACCESS_KEY'.

### DIFF
--- a/changes/437.feature
+++ b/changes/437.feature
@@ -1,1 +1,1 @@
-Add a environment variable `BACKENDAI_ACCESSKEY` for identifying the owner of the session
+Add a environment variable `BACKENDAI_ACCESS_KEY` for identifying the owner of the session
--- a/changes/437.feature
+++ b/changes/437.feature
@@ -1,1 +1,1 @@
-Add a environment variable `BACKENDAI_ACCESS_KEY` for identifying the owner of the session
+Add a environment variable `BACKENDAI_ACCESS_KEY` for identifying the session owner inside the session containers
--- a/changes/437.feature
+++ b/changes/437.feature
@@ -1,0 +1,1 @@
+Add a environment variable `BACKENDAI_ACCESSKEY` for identifying the owner of the session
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -1202,6 +1202,7 @@ async def start_session(
                 ",".join(f"{k}:{v}" for k, v in replicas.items()),
             'BACKENDAI_CLUSTER_HOSTS':
                 ",".join(binding.kernel.cluster_hostname for binding in kernel_agent_bindings),
+            'BACKENDAI_ACCESS_KEY': scheduled_session.access_key
```
adrysn, line 1205: Please append a trailing comma for better diff readability.

src/ai/backend/manager/registry.py
```diff
@@ -1354,6 +1355,7 @@ async def _create_kernels_in_one_agent(
                                 'BACKENDAI_CLUSTER_ROLE': binding.kernel.cluster_role,
                                 'BACKENDAI_CLUSTER_IDX': str(binding.kernel.cluster_idx),
                                 'BACKENDAI_CLUSTER_HOST': str(binding.kernel.cluster_hostname),
+                                'BACKENDAI_ACCESS_KEY': scheduled_session.access_key
```
adrysn, line 1358: Same here.

changes/437.feature
```diff
@@ -0,0 +1 @@
+Add a environment variable `BACKENDAI_ACCESSKEY` for identifying the owner of the session
```
adrysn, line 1: `BACKENDAI_ACCESSKEY` -> `BACKENDAI_ACCESS_KEY`

adrysn (changes requested):
Just a minor request for an update.

achimnol (changes requested):
Setting the env-vars in `start_session()` applies to all kernels (containers) in the same (cluster) session (e.g., `cluster_size`), while doing so in `create_kernels_in_one_agent()` applies to individual kernels (e.g., `cluster_idx`). So the second assignment of `BACKENDAI_ACCESS_KEY` is a duplicate. Please remove it.

PR #440, opened 2021-06-22
fix: prevent iteratively trying to reconnect to redis after restart

leksikov:
OP issue #1245
How to reproduce:

Ensure both Manager and Redis are running normally.
Restart Redis server (container).
Manager displays Can not unset lock "timer.manager_ ... infinitely and unable to re-connect to the Redis.
This spoils the stability of the high-availability configuration of multiple Redis services since there should be some time window to elect a new main Redis node when the previous main node is down. Let's fix this and let the Manager can re-connect when Redis is up again.


leksikov:
My current solution catches the aioredlock redis error in exception. Then destroys the lock manager which was not able to connect to Redis.
When the Redis server alive manager creates a new lock manager as usual.

adrysn:
This PR displays fewer ` Can not set/unset lock ...` error messages by sleeping for the duration of the interval of timers, but it does not resolve the root cause of the problem: re-create a Redis connection pool after Redis connection is down. Perhaps, this problem will automatically be resolved when aioredis 2.0, which supports Redis cluster, is released. Until then, let's look for an alternative way.

leksikov:
One of solution is to implement Redis Watcher. Which will monitor the existence of Redis connection. And if Redis is disconnected then prevent from setting lock variable on disconnected server.

adrysn:
That will be handled by the watcher repository: 
https://github.com/lablup/backend.ai-watcher/issues/2.

But, this is rather a big issue, so it is not feasible to apply now. Also, there should be a way for a Manager to recreate the Redis pool and reconnect to the new when Redis is down and up again. Using Watcher may be useful to monitor the Redis and suppress the lock error, but it does not automatically recreate the pool.

Unfortunately, I don’t have a doable idea on this issue right now. Let’s think and experiment more about this issue.

achimnol:
Closing this PR, because this is actually the responsibility of underlying aioredis library.
If the library does not handle reconnection properly, we may need to add a shim layer between us and the library, or contribute to the library itself.
Let's test out aioredis v2 first.

PR #441, opened 2021-06-27
fix: do not collect data for utilization idle checker when interval is not passed

adrysn:
This PR adds logic not to collect data points for utilization idle checker when the current time does not exceed the sum of the last collect time and the `interval` of the checker. Previously, each Manager process independently collects data without considering the interval, which results in a much faster session termination than intended. 

For example, the idle timeout is reduced by half if there are two Managers since they collect the data twice within the configured interval. By the same logic, if there are three Managers, the idle timeout is reduced by one-third.

So, this PR logs the last collect time and does not collect the data when the `interval` is not passed from it.

### DIFF
--- a/changes/441.fix
+++ b/changes/441.fix
@@ -1,0 +1,1 @@
+Do not collect the data for utilization idle checker when the current time does not exceed the sum of the last collect time and the interval of the checker.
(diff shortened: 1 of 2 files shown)


achimnol (approved):
Good catch, and LGTM.

PR #442, opened 2021-07-05
release: Add API to get available resources

romanceformoon:
lablup/backend.ai#249

This PR adds API to get available resources in groups.
Users can get available resources by API GET method.

### DIFF
--- a/src/ai/backend/manager/api/resource.py
+++ b/src/ai/backend/manager/api/resource.py
@@ -274,6 +274,114 @@ async def check_presets(request: web.Request, params: Any) -> web.Response:
     return web.json_response(resp, status=200)
 
 
+@atomic
+@server_status_required(READ_ALLOWED)
+@auth_required
+@check_api_params(
+    t.Dict({
+        t.Key('scaling_group', default=None): t.Null | t.String,
+        t.Key('group', default='default'): t.String,
+    }))
+async def get_available_resources(request: web.Request, params: Any) -> web.Response:
+    """
+    Returns the list of specific group's available resources.
+
+    :param scaling_group: If not None, get available resources of specific scaling group.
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/api/resource.py
```diff
@@ -748,6 +856,7 @@ def create_app(default_cors_options: CORSOptions) -> Tuple[web.Application, Iter
     add_route = app.router.add_route
     cors.add(add_route('GET',  '/presets', list_presets))
     cors.add(add_route('POST', '/check-presets', check_presets))
+    cors.add(add_route('GET', '/get-resources', get_available_resources))
```
achimnol, line 859: Let's rename the URL to `/available-resources` since the HTTP method name "GET" already indicates its a query API.

achimnol (commented):
* Let's refactor the common reusable parts from `get_available_resources()` and the above existing `check_presets()` handler as a shared function in the `ai.backend.models.resource_preset` module, just like `query_allowed_sgroups()`.
* Though it seems that group resource checks are not implemented yet, please apply [the group resource visibility option](https://github.com/lablup/backend.ai-manager/blob/5a4c0a01db/src/ai/backend/manager/api/resource.py#L258-L260) when you do.

src/ai/backend/manager/models/resource_preset.py
```diff
@@ -193,3 +212,88 @@ async def mutate(
             .where(resource_presets.c.name == name)
         )
         return await simple_db_mutate(cls, info.context, delete_query)
+
+
+async def get_row(conn, request, params, domain_name):
```
achimnol, line 217: Please clarify the name of this function. What is the `row`? Where is it from?

achimnol (changes requested):
Please add type annotations as much as possible when you write new functions.

PR #443, opened 2021-07-06
feat: Add a human readable error message for missing scaling groups

romanceformoon:
lablup/backend.ai#270

Make a error message when raise IntegrityError due to missing scaling groups.

### DIFF
--- a/changes/443.feature
+++ b/changes/443.feature
@@ -1,0 +1,1 @@
+Make a error message when raise IntegrityError due to missing scaling groups.
--- a/src/ai/backend/manager/registry.py
+++ b/src/ai/backend/manager/registry.py
@@ -2215,7 +2215,7 @@ class AgentRegistry:
             agent_info['resource_slots'].items()})
         current_addr = agent_info['addr']
         sgroup = agent_info.get('scaling_group', 'default')
-
+        sent_time = agent_info['time']
         async with self.heartbeat_lock:
 
             instance_rejoin = False
@@ -2307,8 +2307,9 @@ class AgentRegistry:
             try:
                 await execute_with_retry(_update)
             except sa.exc.IntegrityError:
-                log.error(f'Scaling group named [{sgroup}] does not exist.')
-                await asyncio.sleep(5.0)
+                # Log only recent heartbeat in 3 seconds.
+                if now.timestamp() - sent_time < 3.0:
+                    log.error(f'Scaling group named [{sgroup}] does not exist.')
 
             if instance_rejoin:
                 await self.event_producer.produce_event(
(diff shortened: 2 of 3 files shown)


changes/443.feature
```diff
@@ -0,0 +1 @@
+Make a error message when raise IntegrityError due to missing scaling groups.
```
achimnol, line 1: Since this is a Markdwon-formatted text line, let's refine it with more details:
```suggestion
Make an explicit error message upon `IntegrityError` due to missing scaling groups when handling agent heartbeats.
```

src/ai/backend/manager/registry.py
```diff
@@ -2304,7 +2304,11 @@ async def _update() -> None:
                     else:
                         log.error('should not reach here! {0}', type(row['status']))
 
-            await execute_with_retry(_update)
+            try:
+                await execute_with_retry(_update)
+            except sa.exc.IntegrityError:
+                log.error(f'Scaling group named [{sgroup}] does not exist.')
+                await asyncio.sleep(5.0)
```
achimnol, line 2311: Could you explain what this sleep is for?

achimnol (changes requested):

src/ai/backend/manager/registry.py
romanceformoon: It is for delaying error message. When error raises without sleep, that makes too many messages.

achimnol: Hm... is the situation that only happens when restarting the manager after a while with keeping agents running? That's because the heartbeat events are accumulated in Redis and then fetched at the maximum speed upon manager restart. It's an expected behavior.

We need to solve that issue in a different way, possibly adding a wall clock timestamp in heartbeats and ignore "too old" heartbeats, or aggregate multiple heartbeats from same agents into one (leaving only the last one) before processing it.
I'm not sure how to achieve this; would it better to put such "burst-control" feature in the event bus or just implement in the heartbeat handler?

achimnol (changes requested):
I have bad feeling about adding an arbitrary wall-clock based timeout as 3 seconds.
Considering the nature of asynchrony, it is unavoidable to have stale messages upon daemon restarts, and considering the nature of distributed managers, it is impractical to completely eliminate duplicate processing of accumulated burst of heartbeats from the same agents.

I think we need to discuss and think about the design of event-bus burst control in depth, and let's treat it as a separate work.
Please remove the wall-clock timestamping in heartbeats and just print the human-friendly error logs regardless of its rate in this PR.

romanceformoon:
I've selected a CodeBase Study topic which is about heartbeat and event-bus, and i think it will be a good chance to know about that systems deeply. So, I will study about that.

achimnol (approved):
LGTM!

PR #446, opened 2021-07-09
feat: update PyJWT to v2 series

YejeeHa:
lablup/backend.ai#253

Change setup.cfg to make the manager be able to use the PyJWT v2.


achimnol (approved):
Normally all PRs should have the news fragment for auto-generating the changelog upon releases, but I'll skip it as this PR is just a simple bump of a dependency.

PR #450, opened 2021-07-15
feat: Add aiomonitor for manager

romanceformoon:
This PR is started by https://op.lablup.ai/projects/backend-ai-server-and-sdk/work_packages/1414/

aiomonitor: https://github.com/aio-libs/aiomonitor
By using aiomonitor, you can monitoring asynchronous tasks and can execute some asynchronous commands in running application.

You can set port to connect with monitor by editing TOML configs.

### DIFF
--- a/src/ai/backend/manager/config.py
+++ b/src/ai/backend/manager/config.py
@@ -288,6 +288,7 @@ manager_local_config_iv = t.Dict({
         t.Key('hide-agents', default=False): t.Bool,
         t.Key('importer-image', default='lablup/importer:manylinux2010'): t.String,
         t.Key('max-wsmsg-size', default=16 * (2**20)): t.ToInt,  # default: 16 MiB
+        t.Key('aiomonitor-port', default=50001): t.Int,
     }).allow_extra('*'),
     t.Key('docker-registry'): t.Dict({  # deprecated in v20.09
         t.Key('ssl-verify', default=True): t.ToBool,
--- a/src/ai/backend/manager/config.py
+++ b/src/ai/backend/manager/config.py
@@ -288,7 +288,7 @@ manager_local_config_iv = t.Dict({
         t.Key('hide-agents', default=False): t.Bool,
         t.Key('importer-image', default='lablup/importer:manylinux2010'): t.String,
         t.Key('max-wsmsg-size', default=16 * (2**20)): t.ToInt,  # default: 16 MiB
-        t.Key('aiomonitor-port', default=50001): t.Int,
+        t.Key('aiomonitor-port', default=50001): t.Int[1:65535],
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/config.py
```diff
@@ -288,6 +288,7 @@
         t.Key('hide-agents', default=False): t.Bool,
         t.Key('importer-image', default='lablup/importer:manylinux2010'): t.String,
         t.Key('max-wsmsg-size', default=16 * (2**20)): t.ToInt,  # default: 16 MiB
+        t.Key('aiomonitor-port', default=50001): t.Int,
```
achimnol, line 291: You could add the min-max constraints to `t.Int`: `t.Int[1:65535]` to specify the exact range of possible TCP port values. Pleaes update on the lablup/backend.ai-agent#295 as well!

src/ai/backend/manager/server.py
```diff
@@ -633,6 +634,13 @@ async def server_main(loop: asyncio.AbstractEventLoop,
                 str(root_ctx.local_config['manager']['ssl-cert']),
                 str(root_ctx.local_config['manager']['ssl-privkey']),
             )
+
+        # Start aiomonitor.
+        # Port is set by config (default=50001).
+        m = aiomonitor.Monitor(loop, port=root_ctx.local_config['manager']['aiomonitor-port'])
+        m.prompt = "monitor (manager) >>> "
+        m.start()
```
achimnol, line 642: It internally spawns a thread, so we need to call `m.close()` explicitly to properly clean up during shutdown, or use its context manager like `with aiomonitor.start_monitor(...)`.

src/ai/backend/manager/server.py
```diff
@@ -633,6 +634,13 @@ async def server_main(loop: asyncio.AbstractEventLoop,
                 str(root_ctx.local_config['manager']['ssl-cert']),
                 str(root_ctx.local_config['manager']['ssl-privkey']),
             )
+
+        # Start aiomonitor.
+        # Port is set by config (default=50001).
+        m = aiomonitor.Monitor(loop, port=root_ctx.local_config['manager']['aiomonitor-port'])
```
achimnol, line 640: It has [an extra port for telnet-based aioconsole](https://github.com/aio-libs/aiomonitor/blob/master/aiomonitor/monitor.py#L75-L76). Please disable it or provide the port number option for it as well.
Personally I'd prefer disable it since we can just use `python -m aiomonitor -p <port>` command to access the monitor console.

achimnol (changes requested):
Good handling of local configuration. :+1:
Just some minor fixes please!

achimnol (approved):
Good!

PR #454, opened 2021-07-19
feat: add lock-related DB connection settings

adrysn:
Add lock-related DB connection settings for better DB stability when the Manager does not release a lock and/or idle for a long time after acquiring a lock.


achimnol (approved):
LGTM!

PR #455, opened 2021-07-19
fix: Regression of session concurrency limit checks

achimnol:
* There was a mistake while introducing DB transaction retries.


lizable (approved):
LGTM

PR #461, opened 2021-07-27
fix: Fix duplicate codes and add duplicate target dir check for `mount_map` when creating sessions

Taewan-P:
Internal Ticket: OP#1432
This PR removes:
* Duplicate codes of checking the validity of `mount_map` argument.

This PR adds:
* Check duplicate alias folder name.
* Check existing folder name as alias name.
* Check empty alias name.


achimnol (approved):
LGTM!

PR #469, opened 2021-08-20
feat: Add optional manually-assigned agent list to session creation API

kmkwon94:
add agent-list command


src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -469,13 +469,33 @@ async def _schedule_single_node_session(
         log_fmt = _log_fmt.get()
         log_args = _log_args.get()
         try:
-            agent_id = scheduler.assign_agent_for_session(candidate_agents, sess_ctx)
-            if agent_id is None:
-                raise InstanceNotAvailable
+            # OP.1415 if configured sess_ctx.agent_id is existed, skip assin_agent_for_session
```
achimnol, line 472:
```suggestion
            # If sess_ctx.agent_id is already set for manual assignment by superadmin, skip assign_agent_for_session().
```
When you refer the OP issue number, use the format "OP#NNN", but in this case you don't have to refer explicitly, because:
 * You can utilize `git blame` and automated history checks in various IDEs
 * The issue is not a very complicated one and does not require sophisticated/long background context to describe other than the code logic itself.

Also please use the correct English grammar and check typo.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -469,13 +469,33 @@ async def _schedule_single_node_session(
...
-                    sched_ctx, agent_db_conn, sgroup_name, agent_id, sess_ctx.requested_slots,
+                query = (
+                    sa.select([agents.c.available_slots])
+                    .select_from(agents)
+                    .where(agents.c.id == agent_id)
                 )
+                available_agent_slots = (await agent_db_conn.execute(query)).scalar()
+                # if pass the available test
+                if available_agent_slots is None:
+                    raise InstanceNotAvailable("There is no available_agent_slots")
```
achimnol, line 487: In the user-facing strings like exception messages, please use plain English when possible, instead of the variable names directly and give the direct context where the exception originated from so that users can determine what to do next (e.g., try another agent).
```suggestion
                    raise InstanceNotAvailable("There is no such agent.")
```

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -559,22 +579,40 @@ async def _schedule_multi_node_session(
...
-                                extra_conds=agent_query_extra_conds,
-                            )
-                            candidate_agents = await _list_agents_by_sgroup(agent_db_conn, sgroup_name)
-
-                    await execute_with_retry(_reserve)
+                    agent_id = AgentId('')
+                    if kernel.agent_id is not None:
+                        agent_id = kernel.agent_id
+                    else:
+                        agent_id = scheduler.assign_agent_for_kernel(candidate_agents, kernel)
```
achimnol, line 586:
```suggestion
                    agent_id: AgentId | None
                    if kernel.agent_id is not None:
                        agent_id = kernel.agent_id
                    else:
                        agent_id = scheduler.assign_agent_for_kernel(candidate_agents, kernel)
                    assert agent_id is not None
```
Instead of assigning "arbitrary" empty value, it's better to utilize static type checker's capability to narrow down the type scope.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -559,22 +579,40 @@ async def _schedule_multi_node_session(
...
+                    query = (
+                        sa.select([agents.c.available_slots])
+                        .select_from(agents)
+                        .where(agents.c.id == agent_id)
+                    )
+                    available_agent_slots = (await agent_db_conn.execute(query)).scalar()
+                    if available_agent_slots is None:
+                        raise InstanceNotAvailable("There is no available_agent_slots")
```
achimnol, line 594: Same here.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -469,13 +469,33 @@ async def _schedule_single_node_session(
...
+                # if pass the available test
+                if available_agent_slots is None:
+                    raise InstanceNotAvailable("There is no available_agent_slots")
+                available_test_pass = False
+                for key in available_agent_slots:
+                    if available_agent_slots[key] >= sess_ctx.requested_slots[key]:
+                        available_test_pass = True
+                        continue
+                    else:
+                        raise InstanceNotAvailable(f"{key} is insufficent.")
```
achimnol, line 494: Please include more contextual information.
```suggestion
                        raise InstanceNotAvailable(f"The resource slot {key} in {agent_id} does not have the enough remaining capacity.")

```

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -559,22 +579,40 @@ async def _schedule_multi_node_session(
...
+                    available_agent_slots = (await agent_db_conn.execute(query)).scalar()
+                    if available_agent_slots is None:
+                        raise InstanceNotAvailable("There is no available_agent_slots")
+                    available_test_pass = False
+                    for key in available_agent_slots:
+                        if available_agent_slots[key] >= kernel.requested_slots[key]:
+                            available_test_pass = True
+                            continue
+                        else:
+                            raise InstanceNotAvailable(f"{key} is insufficent.")
```
achimnol, line 601: Same here.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -559,22 +579,40 @@ async def _schedule_multi_node_session(
...
+                    if available_test_pass:
+                        async def _reserve() -> None:
+                            nonlocal agent_alloc_ctx, candidate_agents
+                            # assert agent_id is not None
```
achimnol, line 605: If you do `assert` above as suggested, then you don't need this line here.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -559,22 +579,40 @@ async def _schedule_multi_node_session(
...
+                            async with agent_db_conn.begin_nested():
+                                agent_alloc_ctx = await _reserve_agent(
+                                    sched_ctx, agent_db_conn,
+                                    sgroup_name, agent_id, kernel.requested_slots,
+                                    extra_conds=agent_query_extra_conds,
+                                )
+                                candidate_agents = await _list_agents_by_sgroup(
+                                    agent_db_conn, sgroup_name
+                                )
```
achimnol, line 614: Please insert a blank line before/after inner functions.

achimnol (changes requested):

achimnol:
And also the test suite is failing because `PendingSession` now has two additional fields. Please update them.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -559,22 +583,45 @@ async def _schedule_multi_node_session(
...
+
+                                agent_alloc_ctx = await _reserve_agent(
+                                    sched_ctx, agent_db_conn,
+                                    sgroup_name, agent_id, kernel.requested_slots,
+                                    extra_conds=agent_query_extra_conds,
+                                )
+
+                                candidate_agents = await _list_agents_by_sgroup(
+                                    agent_db_conn, sgroup_name
+                                )
```
achimnol, line 623: Insert blank lines before and after the inner function, *not inside* the body!

achimnol (changes requested):
The code you've updated so far looks good.
Now, let's add a new test case in `test_scheduler.py` which tests:
 - success when manually assigned agent is available.
 - failure when mnaually assigned agent is non-existent.
 - failure when manually assigned agent has not enough capacity.

achimnol (changes requested):
* Rename the variable `available_test_pass` to `availability_check_passed` to fix adjective-noun relationships and tense. 
* The test cases look like having copy-and-pasted logic codes from the target codes. Tests should call the target codes and check the behavior from inputs and outputs.

achimnol (approved):

PR #473, opened 2021-08-27
refactor: re-arrange the order of alias folder name checking codes

fregataa:
- change the order of the checking logic which verifies alias folder name. add type hint of yielded folder name before checking iteration.


achimnol (changes requested):
Overall looks good, but please rename `changes/473.feature` to `changes/473.fix` as this is a kind of refactoring and fixing the same functionality.

fregataa:
> Overall looks good, but please rename `changes/473.feature` to `changes/473.fix` as this is a kind of refactoring and fixing the same functionality.

I renamed the file. thank you !

achimnol (approved):

PR #477, opened 2021-09-01
feat: Select and update kernel-pull-progress

youngjun0627:
[backend.ai-issue#227](https://github.com/lablup/backend.ai/issues/227)

- Manager Task
 
  - [x]  Spawn a background task when creating a new session, returning the task ID to the client after max_wait seconds.
  - [x] Let the background task relay the kernel_pull_progress events.

- About Background-Task
The bgtask contains asynchronous functions that select progress status from kernels table or update one on it. To ensure synchronization, I use ###Asyncio-lock. UpdateKernelpullprogress function that update progress status on 'kernels' table is responsible for KernelPullprogress Event. And, [Progress Reporter](https://github.com/lablup/backend.ai-manager/blob/85fe578508ad2edb6cd44587b607157a8d43ae76/src/ai/backend/manager/background.py#L38) is updated by the function. So, manager send the progress-reporter to client, and, client display kernel-pull-progress using the reporter.

### DIFF
--- a/src/ai/backend/manager/api/session.py
+++ b/src/ai/backend/manager/api/session.py
@@ -72,6 +72,7 @@ from ai.backend.common.events import (
     SessionStartedEvent,
     SessionSuccessEvent,
     SessionTerminatedEvent,
+    KernelPullProgressEvent
 )
 from ai.backend.common.logging import BraceStyleAdapter
 from ai.backend.common.utils import cancel_tasks, str_to_timedelta
@@ -481,7 +482,6 @@ async def _create(request: web.Request, params: Any) -> web.Response:
             params['bootstrap_script'] = script
 
     try:
-
         kernel_id = await asyncio.shield(root_ctx.registry.enqueue_session(
             session_creation_id,
             params['session_name'], owner_access_key,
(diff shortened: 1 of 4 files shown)


changes/477.feature
```diff
@@ -0,0 +1 @@
+Add background-task to update progress reporter with KernelPullProgressEvent, until kernel-pulling is done.
```
achimnol, line 1: Please utilize the Markdown syntax and more plain English wording styles.
```suggestion
Add a background task to update the progress reporter with `KernelPullProgressEvent` until image pulling is done.
```

src/ai/backend/manager/api/session.py
```diff
@@ -72,6 +72,7 @@
     SessionStartedEvent,
     SessionSuccessEvent,
     SessionTerminatedEvent,
+    KernelPullProgressEvent
```
achimnol, line 75: Though our code style checker does not check this, please add a trailing comma in multi-line lists, tuples, dicts, and function arguments always.
```suggestion
    KernelPullProgressEvent,
```

src/ai/backend/manager/api/session.py
```diff
@@ -508,52 +509,70 @@ async def _create(request: web.Request, params: Any) -> web.Response:
...
+                if result['status'] == KernelStatus.RUNNING:
+                    break
+                reporter.current_progress = progress[0]
+                reporter.total_progress = progress[1]
+                await reporter.update(0)
+            await asyncio.sleep(0.5)
+
+        task_id = await root_ctx.background_task_manager.start(
+                kernelpullprogress,
+                name='kernel-pull-progress'
```
achimnol, line 556:
```suggestion
                name='kernel-pull-progress',
```

src/ai/backend/manager/api/session.py
```diff
@@ -85,6 +86,7 @@
 )
 from ai.backend.common.plugin.monitor import GAUGE
 
+from ..background import ProgressReporter
```
achimnol, line 89: Tip: This is not a necessary fix, but just FYI. When you don't instantiate a referred type directly but only use it as an annotation, you may put the import inside `if TYPE_CHECKING:` block below, for slightly faster module loads and avoid circular imports by improving independency of each module.

src/ai/backend/manager/api/session.py
```diff
@@ -508,52 +509,70 @@ async def _create(request: web.Request, params: Any) -> web.Response:
...
+                if result is None:
+                    continue
+                if result['status'] == KernelStatus.PREPARING:
+                    await reporter.update(0)
+                if result['status'] == KernelStatus.RUNNING:
+                    break
+                reporter.current_progress = progress[0]
+                reporter.total_progress = progress[1]
+                await reporter.update(0)
+            await asyncio.sleep(0.5)
```
achimnol, line 552: The above loop looks like a busy-wait loop.
Maybe is an indent missing here?

src/ai/backend/manager/api/session.py
```diff
@@ -508,52 +509,70 @@ async def _create(request: web.Request, params: Any) -> web.Response:
...
+        resp['status'] = 'PENDING'
+        resp['servicePorts'] = []
+        resp['created'] = True
+
+        app_ctx.pending_waits.add(current_task)
+        max_wait = params['max_wait_seconds']
+        try:
+            if max_wait > 0:
+                with timeout(max_wait):
+                    await asyncio.sleep(0.5)
```
achimnol, line 569: This breaks the existing API semantics.

src/ai/backend/manager/api/session.py
```diff
@@ -508,52 +509,70 @@ async def _create(request: web.Request, params: Any) -> web.Response:
...
+            else:
+                await asyncio.sleep(0.5)
+
+        except asyncio.TimeoutError:
+            resp['status'] = 'TIMEOUT'
+
```
achimnol, line 575: This breaks the existing API semantics.
Do we really need to remove these to implement the image pulling progress?

achimnol (changes requested):

src/ai/backend/manager/api/session.py
youngjun0627: In the existing API semantics, after enqueue-session is finished, manager sends the response to client. But in this situation, since everything related to create-session is completed and client obtain the response from manager, client does not know bgtask_id during pulling kernel. So, I think that client requests twice, one of them is creating session, the other is getting the information of the existing session.

achimnol: Hmmm. I don't think we need to add another column to the kernels table.

Let's do it as follows:
* When the client calls the session creation API with `enqueue_only` option, the bgtask ID should be included in the response.
  - It may or may not finish soon, but it is the client's job to keep track of session status afterwards.
* When the client calls the session creation API with `max_wait` option:
  - If the creation finishes within the max-wait timeout, just return the full result including service port information and etc. as before. It's usually the case when the image is already populated on the assigned agent(s), and in this case you don't have to reply back the bgtask ID.
  - If not, then include the bgtask ID in the response so that the client could keep track of the image pulling progress.

achimnol: image

youngjun0627: Oh, sorry.. I remove the code to add column of progress on table. I use Only KernelPullProgressEvent! 
I'll try to implement your feedback.

youngjun0627: I have a question.
- Your comment 'I think we may need to keep track of the bgtask ID in the kernels table so that ...' means that we add a column of 'bgtask-id' to 'kernels' table?

achimnol: Ah, sorry for confusion. That's a twist in my head... 🙄  I don't think we need to add bgtask ID to the kernels table.

src/ai/backend/manager/api/session.py
youngjun0627: Thank you, I missed it.

youngjun0627:
I have finished to fix the code reflected your feedback.
And, following images show difference between when max_wait is 0 or not.

- Value of max-wait is equal 0
image

- Set max-wait seconds
image

Also, the code in client-py is fixed. Formally, I don't check 'background_task'. After fixed, I check 'background_task'. If true, kernel-pull-progress is displayed to client during pulling(downloading image). If not, client can't see the progress and wait it.

achimnol:
> I have finished to fix the code reflected your feedback.
> And, following images show difference between when max_wait is 0 or not.
> 
> * Value of max-wait is equal 0
>   image
> * Set max-wait seconds
>   image
> 
> Also, the code in client-py is fixed. Formally, I don't check 'background_task'. After fixed, I check 'background_task'. If true, kernel-pull-progress is displayed to client during pulling(downloading image). If not, client can't see the progress and wait it.

Just a note: Formally -> Formerly
This made me confused for a moment. ;)

achimnol (changes requested):
A few more minor fixes please!

achimnol:
Callable class is generally considered an anti-pattern unless you are writing a kind of low-level library.
Just pass any required information as additional arguments.

achimnol:
To handle multi-node scenarios (cluster sessions) where multiple agents generate the same `KernelPullProgressEvent` with their own progresses, I think there should be a logic to aggregate them and calculate the "total" progress of the multiple agents.

achimnol:
> To handle multi-node scenarios (cluster sessions) where multiple agents generate the same `KernelPullProgressEvent` with their own progresses, I think there should be a logic to aggregate them in the session-level and calculate the "total" progress of the multiple agents. Maybe this could be handled in the client-side, by observing multiple bgtasks concurrently.

To implement this in the client-side, we need to:
* manager: Return the multiple kernel IDs and bgtask IDs for individual kernels from `api.session._create()`
* client-py: Open multiple SSE connections to the bgtask IDs and show the per-kernel progress like multi-layer progress view in Docker CLI's pull command implementation, or aggregate the per-kernel progresses into a single progress bar.

To implement this in the manager-side:
* manager: Return the single bgtask ID just like now.
* manager: In the bgtask, aggregate the multiple kernel pull progress events that belongs to the current session (which may have multiple kernels) and return the aggregated result.

Personally, I think it's better to implement in the client side, as depending on the client-side implementation, we would have runtime options whether to expose more per-kernel details or show the overall progress only.

youngjun0627:
I think the latter is better. So, I'll fix the background task function following your feedback. Then, I'll see client-py source code.

youngjun0627:
- command 1 : `backend.ai run --cluster-mode multi-node --cluster-size 2 'print("hello")' python` 
- command 2 : `backend.ai run -c 'import os; print("Hello world, {}".format(os.environ["CASENO"]))'     -r cpu=1 -r mem=320m     -e 'CASENO=$X'     --env-range=X=case:1,2     python` 
Which one does multi-node scenarios (cluster sessions) mean, command 1 or command 2?

And, if we implement the multi scenario in the client-side, is there no need to fix it anymore?

PR #479, opened 2021-09-16
feat: add start_service API

kyujin-cho:
this commit adds new `start_service` API under session, which starts kernel service on backend.ai session and returns service access information (host and port).


src/ai/backend/manager/api/scaling_group.py
```diff
@@ -27,6 +33,40 @@
     from .context import RootContext
 
 log = BraceStyleAdapter(logging.getLogger(__name__))
+T = TypeVar('T')
+
+
+class DummyHashObject(Generic[T]):
```
achimnol, line 39: The purpose of `DummyHashObject` seems to be for `aiotools.lru_cache` to make a consistent hash value (=cache) for each user group even with different `params` argument in `query_wsproxy_version()` function and to make the dict to be hashable.

I think it would be better to use `dataclasses` or `attrs` to make a hashable attribute set instead of `dict` combined with `DummyHashObject`, in terms of conciseness.

achimnol (changes requested):
Just a minor refactoring please.
Other parts looks good!

achimnol (approved):

PR #480, opened 2021-10-01
feat: Update CRUD of session template.

Sujin-Kim1:
- Modified to create/update multiple session templates at once
- Fields such as `template` or `domain_name` can also be read or updated.
- Correct typo of `example-session-templates.json`

Related: [control-panel#120](https://github.com/lablup/backend.ai-control-panel/pull/120), [webui#1133](https://github.com/lablup/backend.ai-webui/pull/1133)


adrysn (approved):
I tested and it works. LGTM for me.

achimnol (approved):

achimnol (changes requested):
Please check whether checking the emptiness of query results is actually executed or not.
I think it should be rewritten to count the number of returned rows instead of performing Non-check inside the for-loop against the result.

PR #491, opened 2021-10-22
feat: return version information in status API

adrysn:


achimnol (approved):
Actually, the same information is [available via the URL `/`](https://github.com/lablup/backend.ai-manager/blob/6af1d7819e10309cb99a31a410dccc946d7f63be/src/ai/backend/manager/server.py#L146-L153).
Though, it would be better to reduce extra round-trip to the server by adding this to the manager status API.

PR #497, opened 2021-11-12
fix: Remove reusable caching for RPC peer objects

achimnol:
* This will increase overheads for individual RPC calls,
  but may prevent hanging RPC call issues.


achimnol:
I'll merge this PR as it seems to stabilize some customer sites better.

PR #498, opened 2021-11-22
feat: Add "mgr clear-history" manager cli command

yujung7768903:
Add a manager cli command 
* Delete old records from the `kernels` table 
* Invoke the PostgreSQL's vaccuum operation to clear up the disk space.

#### Option
* **'-r', '--retention'**, type=str, default='1yr'
The retention limit.
If set "1yr", it will delete rows from one year ago.
If set "1mo", it will delete rows from one month ago.

* **'-v', '--vacuum-full'**, type=bool, default=False,
If not set, or set False it will run plain VACUUM.
Plain VACUUM(without FULL) simply reclaims space and makes it available for re-use.
If set True, it will run VACUUM FULL.
VACUUM FULL reclaims more space and return a new copy of only the available data. This form is **much slower** and requires an **exclusive lock on each table** while it is being processed. 

#### Screen shot
Screen Shot 2021-11-22 at 11 42 16 AM


src/ai/backend/manager/cli/__main__.py
```diff
@@ -109,6 +113,64 @@ def generate_keypair(cli_ctx: CLIContext):
...
+            expiration_date = today - relativedelta(months=mo)
+        else:
+            duration = TimeDuration()
+            expiration_date = today - duration.check_and_return(retention)
+
+        conn = psycopg2.connect(host=local_config['db']['addr'][0],
+                                port=local_config['db']['addr'][1],
+                                dbname=local_config['db']['name'],
+                                user=local_config['db']['user'],
+                                password=local_config['db']['password'])
```
achimnol, line 149: Just a minor style fix please:
```suggestion
        conn = psycopg2.connect(
            host=local_config['db']['addr'][0],
            port=local_config['db']['addr'][1],
            dbname=local_config['db']['name'],
            user=local_config['db']['user'],
            password=local_config['db']['password'],
        )
```

src/ai/backend/manager/cli/__main__.py
```diff
@@ -7,13 +7,17 @@
...
+import psycopg2
 
 from ai.backend.common.cli import LazyGroup
 from ai.backend.common.logging import BraceStyleAdapter
+from ai.backend.common.validators import TimeDuration
 
 from ..config import load as load_config
 from ..models.keypair import generate_keypair as _gen_keypair
 from .context import CLIContext, init_logger
+from datetime import datetime
```
achimnol, line 19: Please move this import after line 2, since this is an import of the standard library module.

src/ai/backend/manager/cli/__main__.py
```diff
@@ -7,13 +7,17 @@
...
+from dateutil.relativedelta import relativedelta
```
achimnol, line 20: Please move this import after line 9, since this is an import of an external package.

achimnol: Could we replace `relativedelta` with `datetime.timedelta`?

src/ai/backend/manager/cli/__main__.py
```diff
@@ -109,6 +113,64 @@ def generate_keypair(cli_ctx: CLIContext):
...
+    local_config = cli_ctx.local_config
+    with cli_ctx.logger:
+        unit = retention[-2:]
+        today = datetime.now()
+        if unit == 'yr':
+            yr = int(retention[:-2])
+            expiration_date = today - relativedelta(years=yr)
+        elif unit == 'mo':
+            mo = int(retention[:-2])
+            expiration_date = today - relativedelta(months=mo)
```
achimnol, line 140: How about making a new PR to update `ai.backend.common.validators.TimeDuration` to support years and months?

src/ai/backend/manager/cli/__main__.py
```diff
@@ -109,6 +113,64 @@ def generate_keypair(cli_ctx: CLIContext):
...
+            curs.execute(vacuum_sql)
+            conn.set_isolation_level(psycopg2.extensions.ISOLATION_LEVEL_READ_COMMITTED)
+
+            curs.execute("""
+            SELECT COUNT(*) FROM kernels;
+            """)
+            table_size = curs.fetchone()
+            log.info(f'kernels table size: {table_size[0]}')
+
+        log.info('Clear up operation is done.')
```
achimnol, line 171: Let's explicitly print out the calculated expiration date and the number of removed items by counting them before deletion, e.g.,
```suggestion
        log.info('Cleaned up {:,} database records older than {:%Y-%m-%d %H:%M:%S}.', deleted_count, expiration_date)
```

achimnol (changes requested):
Some polishing please!

src/ai/backend/manager/cli/__main__.py
yujung7768903: There are no months and years parameters in timedelta object. So In case of years, `n * 365` days is replaceable. But in case of months, it is difficult to change because it varies from month to month. What should I do?

src/ai/backend/manager/cli/__main__.py
yujung7768903: Okay, I'll do that.

src/ai/backend/manager/cli/__main__.py
achimnol: Hm.. What I wanted here is to unify the return type, but `relativedelta` and `timedelta` looks mostly compatible in terms of arithmetic operators and `relativedelta` has its own custom logic to handle variance of individual month sizes when calculating years and months.
Let's just leave as you first wrote. Sorry for confusion.

yujung7768903: Okay. Then, do I not need to update TimeDuration?

achimnol: I will merge lablup/backend.ai-common#99 soon, then please update this PR to rely on it.

achimnol: Merged it.

achimnol (approved):
Great work!

PR #499, opened 2021-12-03
fix: Adopt aioredis v2 lock (follow-up to #482, #483)

achimnol:
- fix: Rewrite GlobalTimer to use aioredis v2's lock impl.
- fix: Oops
- fix: explicitly extend the lock


achimnol:
Currently on hold due to upstream issues.
The lock is not working as expected...

achimnol:
Closing in favor of etcetra-based locking.

PR #500, opened 2021-12-03
feat: add mutation query for agents table to update schedulable column

lizable:
This PR provides the mutation query of agent setting.
For now, the user with superadmin role can update `schedulable` value only.

Related PR: [backend.ai-webui#1167](https://github.com/lablup/backend.ai-webui/pull/1167).

### DIFF
--- a/src/ai/backend/manager/models/agent.py
+++ b/src/ai/backend/manager/models/agent.py
@@ -324,7 +324,7 @@ class Agent(graphene.ObjectType):
 
 
 class ModifyAgentInput(graphene.InputObjectType):
-    schedulable = graphene.Boolean(required=True, default=True)
+    schedulable = graphene.Boolean(required=False, default=True)
 
 
 class AgentList(graphene.ObjectType):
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/agent.py
```diff
@@ -317,6 +323,10 @@ async def batch_load(
             )
 
 
+class ModifyAgentInput(graphene.InputObjectType):
+    schedulable = graphene.Boolean(required=True, default=True)
```
adrysn, line 327: Let's make this field optional. Other fields of the `agents` table should be updated without altering the value of `schedulable`. For example, I want to add another mutable field, `scaling_group` here, in a separate issue in the near future.

adrysn (changes requested):
Let's make `schedulable` input optional.

adrysn (approved):
LGTM. @achimnol, please double-check and merge if you are okay.

achimnol (approved):
LGTM!

PR #504, opened 2021-12-15
feat: Move max_containers_per_session policy check from predicates

yujung7768903:
Resolve lablup/backend.ai/issues/308

* Move the check for "max-containers-per-session" of keypair resource policies from predicates to enqueueu_session of registry.py
* Regardless of the end of existing sessions , This check is determined by resource policy. 

### Result
* When the number of containers exceeds 'max_containers_per_session'
Screen Shot 2021-12-15 at 8 51 43 PM

### DIFF
--- a/src/ai/backend/manager/api/session.py
+++ b/src/ai/backend/manager/api/session.py
@@ -269,7 +269,7 @@ async def _query_userinfo(
     request: web.Request,
     params: Any,
     conn: SAConnection,
-) -> Tuple[uuid.UUID, uuid.UUID, str]:
+) -> Tuple[uuid.UUID, uuid.UUID, dict]:
     if params['domain'] is None:
         params['domain'] = request['user']['domain_name']
     scopes_param = {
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -757,7 +758,7 @@ async def enqueue_session(
         kernel_enqueue_configs: List[KernelEnqueueingConfig],
         scaling_group: Optional[str],
         session_type: SessionTypes,
-        resource_policy: str,
+        resource_policy: dict,
```
achimnol, line 761: Good catch! 👍🏼

src/ai/backend/manager/api/session.py
```diff
@@ -269,7 +269,7 @@ async def _query_userinfo(
     request: web.Request,
     params: Any,
     conn: SAConnection,
-) -> Tuple[uuid.UUID, uuid.UUID, str]:
+) -> Tuple[uuid.UUID, uuid.UUID, dict]:
```
achimnol, line 272: Good catch! 👍🏼

achimnol (approved):

PR #506, opened 2021-12-19
fix: Avoid TooManySessions error when there is an exact match, not only prefix matches

mindgitrwx:
This pull request temporary solve the session_name matching problem (Too many sessions matched)

https://github.com/lablup/backend.ai/issues/297


mindgitrwx:
https://github.com/lablup/backend.ai/issues/297
I tested this issue with this pull request, And this type of error didn't happen.

achimnol (approved):

PR #507, opened 2021-12-23
Add pipeline data models

achimnol:
This PR adds data models as ORM classes and alembic migrations.


achimnol:
On hold until we resolve lablup/backend.ai#325 to make `pipeline_task` and `session` rows could refer each other via one-to-one foreign keys.

achimnol:
Closing as we are doing this as a separate project.

PR #523, opened 2022-01-21
feat: Limit session types per scaling group

vesselofgod:
resolves https://github.com/lablup/backend.ai/issues/343
Related to: https://github.com/lablup/backend.ai-webui/issues/1190

**Limit session types per scaling group**
- **Check allowed_session_type in scheduled_opts and don't start forbidden type's session**
If it is not a allowed_session_type in scaling_group, do not start the session and stay preparing status
session 안돌게끔

- **Add a column StructuredJSONBColumn to check scheduler_opts's validation using trafaret.**
For validation check, modify "process_result_value" method (receive a result-row column value to be converted)
validation check


src/ai/backend/manager/models/base.py
```diff
@@ -191,6 +194,32 @@ def copy(self):
...
+    def process_bind_param(self, value, dialect):
+        return value if value is not None else None
+
+    def process_result_value(self, raw_value, dialect):
+        # legacy handling
+        schedulerOption = t.Dict(
+            {
+                tx.AliasedKey(['allowed_session_type', 'session_type', 'sessionType'],
+                                default='interactive') >> 'allowed_session_type': tx.Enum(SessionTypes),
+            }).allow_extra('*')
```
achimnol, line 214: This should be provided at the time of instantiation (i.e., the model field definitions), not hard-coded inside the class.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -731,16 +731,28 @@ async def _transition() -> Sequence[PendingSession]:
...
+                        sgroup_name = scheduled_session.scaling_group
+                        async with self.db.begin() as conn:
+                            query = (
+                                sa.select(scaling_groups.c.scheduler_opts)
+                                .select_from(scaling_groups)
+                                .where(scaling_groups.c.name == sgroup_name)
+                            )
+                            result = await conn.execute(query)
+                            row = result.first()
+                            allowed_session_type = row['scheduler_opts']['allowed_session_type']
```
achimnol, line 743: The location of this check should be moved to PENDING -> SCHEDULED transition handler, not here in SCHEDULED -> PREPARING transition handler, to complete the decision **before** proceeding to the actual preparation step.

achimnol (changes requested):

src/ai/backend/manager/models/base.py
vesselofgod: At first, I received the trafaret object as a parameter of StructuredJSONBColumn, but when I tested it, an error occurred
dialector
dialector2
.

achimnol: If you pass a trafaret object when **instantiating** the column class, it should work as expected.

Example:
```python
    sa.Column('scheduler_opts', StructuredJSONBColumn(
        t.Dict({
            t.Key('allowed_session_types', default=['interactive', 'batch']):
                t.List(tx.Enum(SessionTypes)),
        }).allow_extra('*'),
    ), nullable=False, default={}),
```

achimnol: Also there is no need to use `tx.AliasedKey` because it is for legacy clients which use different names for parameter keys.

achimnol:
I've implemented the `StructuredJSONBColumn`. There are more things to fix up and it's up to you! :wink:

src/ai/backend/manager/models/base.py
vesselofgod: When I edit source code, I referred to task_template_v1 in the session_template.py, but I didn't fix it according to the allowed_session_types...😥

src/ai/backend/manager/scheduler/dispatcher.py
vesselofgod: I got it!

achimnol (changes requested):
Just a minor style fix please. :wink:

achimnol (changes requested):
Hm.. I think the check should be implemented as part of the `check_scaling_group` predicate.
This has the following advantages over the current implementation:
 * The user becomes able to know why the session is pending as the reason can be recorded as a part of predicate check results.
 * There no need to issue a separate database transaction. (Note: currently the newly added transaction uses `db.begin()` but it should use `db.begin_readonly()` because it contains only a single SELECT query.)

achimnol (changes requested):
Please move the additional query inside `_query()` so that it could be retried.

achimnol (approved):
Yes, it's a good idea to use already available information. :wink:

PR #524, opened 2022-01-21
feat: Migrate filebrowsers to storage proxy

leksikov:
Add new APIs to map the new manager-facing storage-proxy APIs for file browser.


leksikov:
Implemented the API to facilitate communication between client and Storage Proxy for file browser session. 
There is an issue to think about. When user add his vfolder names how the manager can know which one belongs to which host and volune? Maybe user should also supply host and volume information?

achimnol (changes requested):
`manager/server.py` seems to be passed through black & isort.
Currently we are not modifying existing codebases at once along with a specific feature PR, so please make a separate PR to update code styles and revert the changes caused by black and isort in this PR.
I'd recommend using project-specific (venv-specific) lint configurations if you are using a global configuration.

PR #527, opened 2022-01-28
Add an option to auto-cancel a pending session after specified timeout

vesselofgod:
resolves: https://github.com/lablup/backend.ai/issues/337
Related to: https://github.com/lablup/backend.ai/issues/343

**Add an option to auto cancel a pending session after specified timeout**
- **Get pending_timeout option value to the 'scheduler_opt' and if it's not zero, compare pending_timeout with duration in Pending**
- **Get created_at value to the 'PendingSession' and calculate status duration in Pending**

pending_timeout PR

- **Add a pending_timeout option in scheduler_opt using trafaret**
structured


achimnol:
Please merge the main branch to resolve a conflict with addition of `StructuredJSONBColumn` in both #523 and here.

achimnol (changes requested):
Could you write a test case by setting `created_at` to a time in the past?

vesselofgod:
> you write a test case by setting `created_at` to a time in

ok! I got it!

achimnol:
> > you write a test case by setting `created_at` to a time in
> 
> ok! I got it!

What I meant is to create a new test case with a specific created_at time set in the past (maybe subtract from the current timestamp) and test actual status transition, not just changing the values of timestamps in the test fixtures...

vesselofgod:
> > > you write a test case by setting `created_at` to a time in
> > 
> > 
> > ok! I got it!
> 
> What I meant is to create a new test case with a specific created_at time set in the past (maybe by subtracting from the current timestamp) and test actual status transition, not just changing the values of timestamps in the test fixtures...

I'll add a new test case within this week :)

vesselofgod:
I tried to add a test cases and commit, but error occurred.
I think it's a branch related problem, so I'll solve it and commit.
b

achimnol:
Now need to perform an integratino test.

PR #530, opened 2022-02-03
fix: admins cannot terminate inactivated user's compute sessions

jil8885:
Resolve: https://github.com/lablup/backend.ai/issues/358

Add query code when activated user is not found while remove user session.

### DIFF
--- a/changes/530.fix
+++ b/changes/530.fix
@@ -1,0 +1,1 @@
+Fix admin can't session which deactivated user owns
--- a/changes/530.fix
+++ b/changes/530.fix
@@ -1,1 +1,1 @@
-Fix admin can't session which deactivated user owns
+Allow admins to take actions on behalf of "inactivated" keypairs, such as terminating an RUNNING compute session created by the inactive keypair.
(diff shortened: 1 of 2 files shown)


achimnol (changes requested):
It seems that your code simply overrides the owner access key with the requester access key if the matching keypair is not found.
I think we just need to remove the condition for `keypairs.c.is_active == true()` when querying owner access key **if** the requester is admin or superadmin.

@adrysn How about your opinion?

adrysn (commented):
I think the current implementation can be re-written with fewer lines of code without hitting DB. However, I also do not feel good about implicitly overriding the owner's information with the requester's one. Furthermore, the code just overrides the owner's domain and role to bypass the authorization check logic, but not the owner access key. As a result, the function returns the inactivated owner access key as is. To achieve the same return values, it seems like @achimnol's suggestion is more succinct.

For the @achimnol's comment, the only issue that I can think of for removing `keypairs.c.is_active == true()` is, will it be okay for (super)admins to execute an API on behalf of "inactivated" keypairs. It might not be a problem since the requester is an admin anyway, but it is a little counter-intuitive for me an inactivated keypair can be used to execute a function.

My original idea was just to let the admins destroy the session regardless of the existence of the owner access key, without touching the `get_access_key_scopes`. By doing so, however, we need to modify every API handler which refers to the `get_access_key_scopes`. This may include getting container logs by an admin, etc. This is a bad idea on a second thought.

Since I have no better idea, I agree with @achimnol's suggestion of removing `keypairs.c.is_active == true()` condition.

PR #531, opened 2022-02-08
fix: occupied slot not released when session creation fails

kyujin-cho:
This PR closes lablup/backend.ai#315, by forcing resource usage recalculation when session creation is failed.


adrysn:
Tested and it works well!

achimnol:
Since this is a bug fix, I'll backport it to 21.03 as well.

PR #532, opened 2022-02-14
refactor: store stat data on redis only

kyujin-cho:
This PR resolves lablup/backend.ai#325.

### DIFF
--- a/src/ai/backend/manager/api/resource.py
+++ b/src/ai/backend/manager/api/resource.py
@@ -9,6 +9,7 @@ from decimal import Decimal
 import functools
 import json
 import logging
+import msgpack
 import re
 from typing import (
     Any,
@@ -21,13 +22,15 @@ from typing import (
 import aiohttp
 from aiohttp import web
 import aiohttp_cors
+from aioredis import Redis
+from aioredis.client import Pipeline as RedisPipeline
 from async_timeout import timeout as _timeout
 from dateutil.tz import tzutc
 import sqlalchemy as sa
 import trafaret as t
 import yarl
 
-from ai.backend.common import validators as tx
+from ai.backend.common import redis, validators as tx
 from ai.backend.common.logging import BraceStyleAdapter
 from ai.backend.common.utils import nmget
 from ai.backend.common.types import DefaultForUnspecified, ResourceSlot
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/api/resource.py
```diff
@@ -9,6 +9,7 @@
 import functools
 import json
 import logging
+import msgpack
```
achimnol, line 12: Move this to the next import section as it's a 3rd-party lib. :wink:

src/ai/backend/manager/api/resource.py
```diff
@@ -311,12 +314,23 @@ async def get_container_stats_for_period(request: web.Request, start_date, end_d
         result = await conn.execute(query)
         rows = result.fetchall()
 
+    async def _pipe_builder(r: Redis) -> RedisPipeline:
+        pipe = r.pipeline()
+        for row in rows:
+            pipe.get(str(row['id']))
+        return await pipe.execute()
```
achimnol, line 321: Just return the pipe using a synchronous function. ([ref](https://github.com/lablup/backend.ai-common/blob/main/tests/redis/test_pipeline.py))

achimnol (changes requested):
Just two minor code issues.

src/ai/backend/manager/api/resource.py
```diff
@@ -311,12 +314,23 @@ async def get_container_stats_for_period(request: web.Request, start_date, end_d
         result = await conn.execute(query)
         rows = result.fetchall()
 
+    async def _pipe_builder(r: Redis) -> RedisPipeline:
```
fregataa, line 317: 굉장히 사소하긴 하지만, 여기 async가 없어도 되나요?

fregataa (commented):
leave a minor comment.

src/ai/backend/manager/api/resource.py
kyujin-cho: good catch :+1:

achimnol (changes requested):
`common.redis.execute()` accepts a function that returns either:
 - async function that executes Redis commands
 - non-async function that returns a Redis pipeline to execute
To use pipeline, the callable must *NOT* be async.

achimnol:
> `common.redis.execute()` accepts a function that returns either:
> 
> * async function that executes Redis commands
> * non-async function that returns a Redis pipeline to execute
> 
> To use pipeline, the callable must _NOT_ be async.
> 
> refs: https://github.com/lablup/backend.ai-common/blob/3a6466fe/src/ai/backend/common/redis.py#L228-L234

I just added a PR to allow async pipeline builders: lablup/backend.ai-common#125.

achimnol (changes requested):
I think it's better to merge `redis clear-history` with `clear-history` command.  How about your thoughts?

achimnol (approved):

PR #536, opened 2022-02-17
feat: detach keypair concurrency table

fregataa:
the implementation of 2nd approach in https://github.com/lablup/backend.ai/issues/365

# Test
## TL; DR
Migrating concurrency check to redis works without malfunctions. Other tables such as `kernels` causes a bottleneck but this is another issue.
Creating a `keypair_resource_usages` table occurs many bottlenecks and consequently raises canceled error.

## 1
```
python run.py start --cpu=0.5 --mem=0.5g -n 100 -d 2 --max-parallel=10 sleep
```
The number of sessions is too big to examine the change.

A bottleneck occurs on the created table `keypair_resource_usages`.
image

## 2
```
python run.py start --cpu=0.5 --mem=0.5g -n 20 -d 0.1 --max-parallel=16 sleep
```

### Result
A deadlock occurs. One session does not finish its task.
image
image

### Error

Bottlenecks on DB occur.
image
image

## 3
```
python run.py start --cpu=0.5 --mem=0.5g -n 20 -d 0.1 --max-parallel=32 sleep
```

### Result
image

image

### Error
image
image


achimnol (changes requested):
Though it is still draft, here are a quick early review:
 * Remove `dev.etcd.volumes.json` which looks like accidentally added.
 * Let's rename the `keypair_concurrency` to `keypair_resource_usage` for cases when we could add additional fields.

achimnol:
Closing in favor of #535!
Thanks for the work 👍🏼

PR #537, opened 2022-02-17
feat: Allow mounting vfolder subdirectory into session container

SonMinWoo:
resolves: https://github.com/lablup/backend.ai/issues/360
related: https://github.com/lablup/backend.ai-agent/pull/343

**Let's allow using a "partial" view of a vfolder inside sessions.
For example, if a vfolder mystuff has a/b/c subdirectories, let's allow mounting mystuff/a/b as /home/work/bbb.**

When vfolders are mounted, for verification existence of it db query is used. However, that have only top path of subdirectory and make object with necessary information to mount including only top path of subdirectory as folder name.

To solve this problem, after making other information to mount using top path of subdirectory edit only top path of folder names to full folder name including subdirectories. 

issue360


src/ai/backend/manager/registry.py
```diff
@@ -848,25 +852,34 @@ async def enqueue_session(
...
+            raise VFolderNotFound(extra_data=[*(set(root_mounts) - matched_mounts)])
+        tmp_mounts = list()
+        while determined_mounts:
+            mount_obj = determined_mounts.pop()
+            for key in root_mount_map.keys():
+                if mount_obj[0] == key:
+                    for item in root_mount_map[key]:
+                        tmp_obj = copy.deepcopy(mount_obj)
+                        tmp_obj[0] = item
+                        tmp_mounts.append(tmp_obj)
```
achimnol, line 881: These lines could be just replaced with recreating tuples with modified values.

src/ai/backend/manager/registry.py
```diff
@@ -777,6 +777,10 @@ async def enqueue_session(
 
         mounts = kernel_enqueue_configs[0]['creation_config'].get('mounts') or []
         mount_map = kernel_enqueue_configs[0]['creation_config'].get('mount_map') or {}
+        root_mounts = list(set(map(lambda x: x.split("/")[0], mounts)))
+        root_mount_map = dict()
```
achimnol, line 781: For intrinsict container types, it's recommended to use the native syntax (`{}` instead of `dict()`).

Though, using `dict()` may be better if there are set expressions around it to avoid confusion of that `{}` is an empty dict while `{something, ...}` (without `:`) is a set expression.  I think this case is not.

src/ai/backend/manager/registry.py
```diff
@@ -777,6 +777,10 @@ async def enqueue_session(
 
         mounts = kernel_enqueue_configs[0]['creation_config'].get('mounts') or []
         mount_map = kernel_enqueue_configs[0]['creation_config'].get('mount_map') or {}
+        root_mounts = list(set(map(lambda x: x.split("/")[0], mounts)))
```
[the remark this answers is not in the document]
achimnol: Could you also use the list/set comprehension syntax to make it more concise?

src/ai/backend/manager/registry.py
```diff
@@ -777,6 +777,10 @@ async def enqueue_session(
...
+        root_mount_map = dict()
+        for item in root_mounts:
+            root_mount_map[item] = list(filter(lambda x: x.find(item) >= 0, mounts))
```
achimnol, line 783: How about rewriting this using the list comprehension syntax?

achimnol: Could you explain what is the purpose of `root_mount_map` variable?

changes/537.feature
```diff
@@ -0,0 +1 @@
+Check resource by root folder name of subdirectory and apply mount info of root to its mounted subdirectories.
```
achimnol, line 1: What do you mean by "check resource" (e.g., is it related with resource slots)?

achimnol (changes requested):

src/ai/backend/manager/registry.py
```diff
@@ -848,25 +852,34 @@ async def enqueue_session(
                     # the server may respond with error if the directory already exists
                     pass
                 matched_mounts.add(item['name'])
-                determined_mounts.append((
+                determined_mounts.append([
```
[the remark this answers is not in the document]
SonMinWoo: It is to edit list element, changed make new tuple.

src/ai/backend/manager/registry.py
SonMinWoo: `root_mount_map` is not necessary, so it is deleted. Now `mount` play this role.

changes/537.feature
SonMinWoo: This refers to an already implemented function that checks whether the folder to be mounted exists or not. 
And this function also returns various information such as permission, host, path, etc.

So It is implemented carrying all subdirectories' various informations to `backend.ai-agent`, and `backend.ai-agent` check subdirectories' existence (if not exist, don't work and return error message). This feature is already implemented, too.

PR #538, opened 2022-02-18
fix: Let `get_wsproxy_version()` be aware of per-user/per-group scaling group mapping

kyujin-cho:
This PR fixes issue `get_wsproxy_version` returning 404 when target scaling group is allowed to invididual keypair/user group rather than user's associated domain.


achimnol (approved):
LGTM!

PR #539, opened 2022-02-18
Fix/prevent blocking of the whole manager while cloning a vfolder

vesselofgod:
resolved : https://github.com/lablup/backend.ai/issues/354

**prevent blocking of the whole manager while cloning a vfolder**
- **Run the clone operation as a background task and just return the bgtask ID in the vfolder clone API, so that clients could keep track of the progress and result.**
clone pr

- **Split the destination vfolder creation and actual clone operations, and insert the database transaction to create the destination vfolder record between them.**
split

### DIFF
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -1828,6 +1828,7 @@ async def clone(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
             'user': user_uuid,
             'group': group_uuid,
             'cloneable': params['cloneable'],
+            'bgtask_id': task_id.hex,
         }
         query = (sa.insert(vfolders, insert_values))
         try:
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -25,6 +25,7 @@ import aiohttp
 from aiohttp import web
 import aiohttp_cors
 import sqlalchemy as sa
+from ..background import ProgressReporter
 import trafaret as t
 
 from ai.backend.common import validators as tx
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/api/vfolder.py
```diff
@@ -1774,33 +1800,11 @@ async def clone(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
...
-                pass
-        except aiohttp.ClientResponseError:
-            raise VFolderOperationFailed
-
         user_uuid = str(user_uuid)
         group_uuid = None
         ownership_type = 'user'
         insert_values = {
-            'id': folder_id.hex,
+            'id': uuid.uuid4().hex,
```
achimnol, line 1807: Where is the original folder ID?
What happens if the folder ID is lost?

src/ai/backend/manager/api/vfolder.py
```diff
@@ -25,6 +25,7 @@
 from aiohttp import web
 import aiohttp_cors
 import sqlalchemy as sa
+from ..background import ProgressReporter
```
achimnol, line 28: Please reorder the import following other imports.
(What is the rule?)

changes/539.fix
```diff
@@ -0,0 +1 @@
+Split the destination vfolder creation and actual clone operationsrun the clone operation as a background task and just return the bgtask ID in the vfolder clone API, so that clients could keep track of the progress and result
```
achimnol, line 1: Please fix typos and try to write it shorter.

achimnol (changes requested):

src/ai/backend/manager/api/vfolder.py
vesselofgod: Thank you for the good point! I thought that UUID and folder_id would always be the same. But, it was actually different and found that other tasks of vfolder had errors. so, I will fix it.

src/ai/backend/manager/api/vfolder.py
achimnol: Try out finding by yourself. How are the current imports organized?

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1824,6 +1828,7 @@ async def clone(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
             'user': user_uuid,
             'group': group_uuid,
             'cloneable': params['cloneable'],
+            'bgtask_id': task_id.hex,
```
achimnol, line 1831: `bgtask_id` is a UUID field.  Though this still works because `tx.UUID()` also accepts the hexademical form, I think it is better to keep the original UUID format by replacing this with `str(task_id)`.

changes/539.fix
```diff
@@ -0,0 +1 @@
+Split vfolder creation and clone operations and run clone operation as a background task and clients keep track of the progress and result
```
achimnol, line 1:
```suggestion
Prevent potential blocking of mutating database transactions when vfolder clone operations take too long time, by making clone operation async (background task) with transactions
```

achimnol (changes requested):

achimnol:
Just a note for myself.
Need to refactor: move the following lines in `api.vfolder.create()` and `api.vfolder.clone()` to `SharedConfig`:
image

achimnol:
The PR itself also needs to split the txn block.  Currently the storage proxy RPC call has race condition with the target vfolder creation.

PR #542, opened 2022-02-19
fix: Don't allow empty values in allowed session types and scan all allowed scaling groups

vesselofgod:
This is a follow-up to lablup/backend.ai#343
Related with: lablup/backend.ai-webui#1190

**prevent allowed session types in empty and fix allowed session types check process handing**

- **Prevent allowed_session_types in empty status  using trafaret.**
Change it not to be empty to solve problem of If the scaling group's allowed_session_types is empty, predicate check fail failure.
비어있는 경우
비어있는 경우가 없도록 수정

- **Fix allowed_session_types check process handing**
Prevent stopping without checking all sgroups during the allowed_session_types check process

### DIFF
--- a/src/ai/backend/manager/scheduler/predicates.py
+++ b/src/ai/backend/manager/scheduler/predicates.py
@@ -216,6 +216,14 @@ async def check_scaling_group(
     if preferred_sgroup_name is not None:
         for sgroup in sgroups:
             if preferred_sgroup_name == sgroup['name']:
+                allowed_session_types = sgroup['scheduler_opts']['allowed_session_types']
+                if sess_ctx.session_type.value.lower() not in allowed_session_types:
+                    return PredicateResult(
+                        False,
+                        f"The given preferred scaling group is not allowed to use. "
+                        f"({preferred_sgroup_name})",
+                        permanent=True,
+                    )
                 break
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -228,19 +236,21 @@ async def _query():
...
+        for sgroup in sgroups:
+            allowed_session_types = sgroup['scheduler_opts']['allowed_session_types']
+            if sess_ctx.session_type.value.lower() in allowed_session_types:
+                break
+        else:
+            return PredicateResult(
+                False,
+                "Not allowed session type in scaling groups.",
+            )
         target_sgroup_names = [sgroup['name'] for sgroup in sgroups]
```
achimnol, line 248: This logic should be rewritten to exclude the scaling groups that do not allow the desired session type and report failure *only when there is no scaling group left*.

achimnol (changes requested):

src/ai/backend/manager/scheduler/predicates.py
achimnol: Rewrote and add a test case for the behavior.

PR #544, opened 2022-02-21
feat: Allow superadmins to list vfolders as specific user

SonMinWoo:
resolve: lablup/backend.ai#321
related: lablup/backend.ai-webui#1237

To solve the problem of webui, this is implemented in `list_folders` api. This feature uses only the user's email to search the folder of the user. Now only superadmin can use this feature, but it will be expandable like public/private folder option.

It request to DB about vfolder informations belong to carried user email. The other process is same to existing folder listing feature.
issue321-2

When superadmin click checkbox and it is checked, webui request to manager this feature. Parameter is added 'user email'.


achimnol (changes requested):
To keep consistency with `owner_access_key` used in session APIs to delegate the identity, let's rename the new user argument as `owner_user_email`.

achimnol (commented):
Q: Why not used `query_accessible_vfolders()` with the target user UUID and role?

SonMinWoo:
> Q: Why not used `query_accessible_vfolders()` with the target user UUID and role?

A: It is because double query is needed to know the target user UUID and role by only email. I thought this process cause performance degradation.

achimnol:
> > Q: Why not used `query_accessible_vfolders()` with the target user UUID and role?
> 
> A: It is because double query is needed to know the target user UUID and role by only email. I thought this process cause performance degradation.

That overhead is okay because it only fetches one row from a unique key.  The `owner_access_key` (`get_access_key_scopes()`) already doing the same.  I've rewritten so by myself.

src/ai/backend/manager/api/utils.py
```diff
@@ -69,29 +70,86 @@ async def get_access_key_scopes(request: web.Request, params: Any = None) -> Tup
...
+                raise GenericForbidden(
+                    "Domain-admins cannot perform operations on behalf of super-admins.",
+                )
+            pass
+        else:
+            raise GenericForbidden(
+                "Only admins can perform operations on behalf of other users.",
+            )
+    else:
+        owner_user_uuid = request['user']['uuid']
```
lizable, line 143: I fixed this part since there's no `user_uuid` or `user_role` field in the request parameter.
Instead, we can use `uuid` and `role` field from user field in request.

lizable (commented):
I fixed minor things to make `list_folders()` function works fine.

src/ai/backend/manager/api/utils.py
achimnol: Thanks! 👍🏼

achimnol:
Let's deprecate the "all" option as it would incur severe performance overheads when used on the cloud.  I will update the code.

PR #551, opened 2022-03-15
fix: give load aliases arg

fregataa:
give a `load_aliases` argument when it is needed

methods which query select from `image` use `load_aliases` parameter to define eager loading of `aliases` relationship.
results of the query with `load_aliases = False` cannot load `aliases` relationship.


kyujin-cho (approved):
LGTM!

PR #554, opened 2022-03-18
fix: automatic round-down to quantum size

kyujin-cho:
This PR resolves https://github.com/lablup/backend.ai-internal/issues/34.


src/ai/backend/manager/registry.py
```diff
@@ -1273,10 +1272,54 @@ async def _finialize_running() -> None:
...
+                        actual_allocs: MutableMapping[str, str] = {}
+                        for alloc_map in created_info['resource_spec']['allocations'].values():
+                            for slot_name, allocations in alloc_map.items():
+                                total_allocs: List[Decimal] = []
+                                for allocation in allocations.values():
+                                    if BinarySize.suffix_map.get(allocation[-1].lower()) is not None:
+                                        total_allocs.append(Decimal(BinarySize.from_str(allocation)))
+                                    else:
+                                        total_allocs.append(Decimal(allocation))
+                                actual_allocs[slot_name] = str(sum(total_allocs))
```
achimnol, line 1285: Could we extract this as a separate function and add a unit test case?

src/ai/backend/manager/registry.py
```diff
@@ -1273,10 +1272,54 @@ async def _finialize_running() -> None:
...
+
+                        values['occupied_slots'] = actual_allocs
+                        query = (
```
achimnol, line 1288: Let's name the variables that points sqlalchemy queries differently, e.g., `select_query` and `update_query`, etc. (or maybe `read_query`, `info_query`, etc.).
This is to prepare adoption of sqlalchemy typing in the future and avoid conflicts of different types of the same-named variables.

src/ai/backend/manager/registry.py
```diff
@@ -1273,10 +1272,54 @@ async def _finialize_running() -> None:
...
+                        alloc_diffs = {
+                            k: Decimal(requested_slots[k]) - Decimal(v)
+                            for k, v in actual_allocs.items()
+                        }
+                        if any([v != 0 for v in alloc_diffs.values()]):
+                            agent_actual_allocated_slots = {
+                                k: str(Decimal(agent_occupied_slots[k]) - v)
+                                for k, v in alloc_diffs.items()
+                            }
+                            query = (
```
achimnol, line 1309: Let's rename it `update_query`.

src/ai/backend/manager/registry.py
```diff
@@ -1273,10 +1272,54 @@ async def _finialize_running() -> None:
...
+                                .where(agents.c.id == agent_alloc_ctx.agent_id)
+                            )
+                            await conn.execute(query)
+                        query = (
+                            kernels.update()
+                            .values(values)
+                            .where(kernels.c.id == created_info['id']))
+                        await conn.execute(query)
+                except Exception as e:
+                    log.exception('error while executing _finalize_running', exc_info=e)
```
achimnol, line 1321: Inside an `except` clause block, you don't have to explicitly pass `exc_info` as `log.exception()` will automatically pick up `sys.exc_info()`.

src/ai/backend/manager/registry.py
```diff
@@ -1273,10 +1272,54 @@ async def _finialize_running() -> None:
...
+                            agent_actual_allocated_slots = {
+                                k: str(Decimal(agent_occupied_slots[k]) - v)
+                                for k, v in alloc_diffs.items()
+                            }
+                            query = (
+                                agents.update()
+                                .values({'occupied_slots': agent_actual_allocated_slots})
+                                .where(agents.c.id == agent_alloc_ctx.agent_id)
+                            )
+                            await conn.execute(query)
```
achimnol, line 1314: Updating kernel-specific resource occupation here looks good because this is a per-kernel "running" finalization handler.  However, for cluster sessions, this would incur multiple concurrent update transactions to the agents table.  How about aggregating and splitting out the agent table update into a single query using a single transaction after updating individual kernels?

achimnol (changes requested):

src/ai/backend/manager/registry.py
kyujin-cho: In that case, how about just calling `AgentRegistry.recalc_resource_usage()` after gathering all agent tasks?

achimnol (approved):

PR #560, opened 2022-03-28
Fix SQL query errors in image aliases and ensure exhaustive iteration of given reference candidates

fregataa:
1. Any image alias cannot be fetched since `UnknownImageReference` error is raised before iterating all `reference_candidates` in `resolve()` method. Let the error lazy-occur and the iteration finish to resolve all `reference_candidates`.
2. Since classes, queries and other SQL elements are not accepted in the join subquery context, update the query statement in `from_alias()` method.
 
image


achimnol (changes requested):
I'm fixing this as follows:
* Python docstrings should use reStructuredText instead of Markdown.
* Typos in English sentences, such as missing captialization of the first letter.
* Flag-like optional arguments are better to be declared as explicit keyword arguments using `*`. Also it is better to keep their ordering in different methods/functions consistent for readability.
* Make the docstring and code consistent in use of `ImageAlias` and `str`.
* Clarify "what stuff is strict" by renaming `strict` to `strict_arch`.

Things to fix from your side:
* `common.types.ImageAlias` should be exposed as part of `common.types.__all__`. -> Please make another PR in the common.
* "good use case" is very strange terminology because desirably the method should have one single direct purpose.  You should not be confused between the complexity of operation and the mix of operations.

achimnol:
> Things to fix from your side:
> 
> * `common.types.ImageAlias` should be exposed as part of `common.types.__all__`. -> Please make another PR in the common.
> * "good use case" is very strange terminology because desirably the method should have one single direct purpose.  You should not be confused between the complexity of operation and the mix of operations.

@kyujin-cho Please clarify the docstring and the purpose of `resolve()`, as it is first written by you.

achimnol (approved):

PR #561, opened 2022-03-29
fix: refer vfolder_mounts instead of will-be-deprecated mounts in returning session usage statistics

adrysn:
* The structure of the elements of `mounts` is changed from `list` to `str`, so we cannot access it as we did before 22.03.
* To make the contents of NFS value more useful, now it returns the physical host location of vfolders mounted, not just vfolder host, which was not so needed.

### DIFF
--- a/src/ai/backend/manager/api/resource.py
+++ b/src/ai/backend/manager/api/resource.py
@@ -335,7 +335,7 @@ async def get_container_stats_for_period(request: web.Request, start_date, end_d
         if row['vfolder_mounts']:
             # For >=22.03, return used host directories instead of volume host, which is not so useful.
             nfs = list(set([str(mount.host_path) for mount in row['vfolder_mounts']]))
-        elif row['mounts'] and len(row['mounts']) > 0 and isinstance(row['mounts'][0], list):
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/resource.py
```diff
@@ -332,8 +332,12 @@ def _pipe_builder(r: Redis) -> RedisPipeline:
             continue
         last_stat = msgpack.unpackb(raw_stat)
         nfs = None
-        if row['mounts'] is not None:
-            nfs = list(set([mount[1] for mount in row['mounts']]))
+        if row['vfolder_mounts']:
+            # For >=22.03, return used host directories instead of volume host, which is not so useful.
+            nfs = list(set([str(mount.host_path) for mount in row['vfolder_mounts']]))
+        elif row['mounts'] and len(row['mounts']) > 0 and isinstance(row['mounts'][0], list):
```
achimnol, line 338:
```suggestion
        elif row['mounts'] and isinstance(row['mounts'][0], list):
```
You may rely on the Python's truthness semantic about the emptiness of lists as PEP-8 recommends.

achimnol (changes requested):

achimnol (approved):

PR #562, opened 2022-03-30
refactor: final touches on image table

kyujin-cho:
This PR updates `ImageRow.resolve()`'s comment to describe how it works more clearly, fixes some invalidly assigned parameters, and removes logging calls for debug purposes.


kyujin-cho:
This needs to be done at #561.

PR #567, opened 2022-03-31
feat: Session lifecycle event callback as webhook

achimnol:
refs lablup/backend.ai#362


adrysn (changes requested):
The webhook request is successfully received by the MLOps pipeline server. However, the session event order is not guaranteed. For example, in the following Manager log, the `started` event is sent after the `terminated` event:
image

inureyes (approved):
LGTM (bugfix part)

PR #573, opened 2022-04-04
refactor: union vfolders into one query

agatha197:
To successfully support the pagination query, the vfolder list is united into one.
Previously, `query_accessible_vfolders()` was implemented with Python's list concat, but now only one query containing all vfolder lists is returned.


achimnol:
Replaced with lablup/backend.ai#656

PR #574, opened 2022-04-07
fix: rounded down resource not released after kernel creation

kyujin-cho:
This PR fixes `occupied_slots` column on `agents` table not matching to actual resource occupied on corresponding agent when fractional resource requested is rounded down to quantum size when creating kernel on agent.

### DIFF
--- a/src/ai/backend/manager/registry.py
+++ b/src/ai/backend/manager/registry.py
@@ -1588,7 +1588,7 @@ class AgentRegistry:
 
     async def settle_agent_alloc(
         self, kernel_agent_bindings: Sequence[KernelAgentBinding],
-    ):
+    ) -> None:
         """
         Tries to settle down agent row's occupied_slots with real value. This must be called
         after kernel creation is completed, to prevent fraction of resource dropped by agent scheduler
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -1582,6 +1586,54 @@ async def update_scaling_group(self, id, scaling_group) -> None:
         ) as rpc:
             await rpc.call.update_scaling_group(scaling_group)
 
+    async def calibrate_agent_alloc_differences(
+        self, kernel_agent_bindings: Sequence[KernelAgentBinding],
+    ):
+        """
+        Tries to calibrate agent row's occupied_slots with real value. This must be called
```
achimnol, line 1593: Update the comment accordingly.

src/ai/backend/manager/registry.py
```diff
@@ -1582,6 +1586,54 @@ async def update_scaling_group(self, id, scaling_group) -> None:
...
+        after kernel creation is completed, to prevent fraction of resource dropped by agent scheduler
+        during kernel creation still being reported as used.
+        """
+
+        def _str_to_decimal(orig_slots: ResourceSlot):
```
achimnol, line 1598: Specify the return type.
How about renaming to `_numericalize`?
How about making it a method of `ResourceSlot`?

achimnol (changes requested):

src/ai/backend/manager/registry.py
kyujin-cho: Just found that `ResourceSlot.from_json()` converts value explicitly to `Decimal`. I'll remove this inner function and replace usages with `from_json()` calls.

adrysn:
I tested this PR with the latest commit and confirmed that it works fine in a functional aspect:

image

achimnol (approved):

PR #580, opened 2022-04-19
fix: do not query update where update is empty

fregataa:
image

a method `sync_kernel_stats` in ai.backend.manager.registry.AgentRegistry, `per_kernel_updates` is used to bind update query parameters.
`per_kernel_updates` can be empty but it tries query the update with no binded parameter.


achimnol (approved):
👍🏼

PR #583, opened 2022-04-21
feat: Add `KeyPairResourcePolicy.max_session_lifetime`

achimnol:
resolves lablup/backend.ai-internal#46

### DIFF
--- a/src/ai/backend/manager/idle.py
+++ b/src/ai/backend/manager/idle.py
@@ -343,6 +343,67 @@ class TimeoutIdleChecker(BaseIdleChecker):
         return False
 
 
+class SessionLifetimeIdleChecker(BaseIdleChecker):
+
+    name: ClassVar[str] = "session_lifetime"
+    check_interval: float = 10.0
+
+    _policy_cache: ContextVar[Dict[AccessKey, Optional[Mapping[str, Any]]]]
+
+    async def __ainit__(self) -> None:
+        await super().__ainit__()
+        self._policy_cache = ContextVar("_policy_cache")
+
+    async def populate_config(self, config: Mapping[str, Any]) -> None:
+        pass
+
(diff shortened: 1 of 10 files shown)


achimnol:
TODO: add a timer to periodically check expired sessions

src/ai/backend/manager/idle.py
```diff
@@ -343,13 +370,29 @@ async def check_session(self, session: Row, dbconn: SAConnection) -> bool:
...
+
+    async def populate_config(self, config: Mapping[str, Any]) -> None:
+        pass
+
+    async def check_session(self, session: Row, dbconn: SAConnection, policy: Row) -> bool:
+        now = await dbconn.execute(sa.select(sa.func.now()))
+        if policy["max_session_lifetime"] >= 0:
+            # TODO: once per-status time tracking is implemented, let's change created_at
+            #       to the timestamp when the session entered PREPARING status.
+            if now - session["created_at"] >= policy["max_session_lifetime"]:
```
adrysn, line 385: If `max_session_lifetime` is zero, I suggest interpreting it as infinity. Also, these lines raise exceptions due to calculation type-mismatch issues. These might be updated to something like below:

```python
now = await dbconn.scalar(sa.select(sa.func.now()))
if policy["max_session_lifetime"] > 0:
    # TODO: once per-status time tracking is implemented, let's change created_at
    #       to the timestamp when the session entered PREPARING status.
    if now - session["created_at"] >= timedelta(seconds=policy["max_session_lifetime"]):
```

adrysn (changes requested):
I found an error in `SessionlifetimeChecker. Other parts seem to be working well in my test environment.

adrysn (approved):
Tested and LGTM.

PR #587, opened 2022-05-03
add feature audit_logs

hydroxyde:
This pull request refers to: https://github.com/lablup/backend.ai-internal/issues/17

This is the implementation of the audit log feature's model which keeps records of creation, modification and deletion, etc., of users, keypairs and vfolders. 
### Audit Logs model
Each record in the audit log is as follow:

**The subject** (i.e., who did the action?)
It contains: the API access key, user UUID, user email (with respective types, all indexed)

**The action**

- Action type: CREATE | CHANGE | DELETE

- Action data: (JSONB field)

{"before": {"data1": "val1", ...}, "after": {"data1": "val2", ...}}
For passwords and secret keys, we do not record the original/hash values.
CREATE: record only "after", leaving "before" as null
CHANGE: record both "before", "after" and include only the changed data.
DELETE: record only "before", leaving "after" as null

**The target object**
It contains the values of users.uuid, keypairs.access_key or vfolder.name

**The date/time** of action (indexed, this should be the default descending order key of audit log queries)

> Screenshot 2022-05-04 at 17 29 36

### Periodic removal of audit logs
This command extends #498 and gives option to remove audit logs based on time (e.g., remove all logs older than 3 months, 1 year, etc...)

Screenshot 2022-05-10 at 10 25 16

**Extended options are as follow:**

- **-t, --table [kernels|audit_logs]** : Which table to run the operation on. Options available are kernels  and audit_logs. If not set or set to kernels, it will run the operation on kernels table. If set to audit_logs, it will run on audit_logs table (default: kernels). 
- **--target** : For audit_logs you can chose which audit target you would like the operation to run on. You can provide user.uuid, keypairs.access_key or vfolder_name. If not set, it will run on all targets.


hydroxyde:
Retention command needs update as I was first using target column as 'user' or 'keypair' then switched to users.uuid and keypairs.access_key. So I am thinking to remove this option, or would it be worth it to have such implementation ?

achimnol:
Replaced with lablup/backend.ai#502.

PR #588, opened 2022-05-06
fix: add automount folder when session creation occurs

lizable:
This PR resolves the issue raised on [backend.ai-webui#1302](https://github.com/lablup/backend.ai-webui/issues/1302).

### DIFF
--- a/src/ai/backend/manager/models/vfolder.py
+++ b/src/ai/backend/manager/models/vfolder.py
@@ -24,7 +24,7 @@ import trafaret as t
 
 from ai.backend.common.types import VFolderMount
 
-from ..api.exceptions import InvalidAPIParameters, VFolderNotFound, VFolderOperationFailed
+from ..api.exceptions import InvalidAPIParameters, VFolderOperationFailed
 from ..defs import RESERVED_VFOLDER_PATTERNS, RESERVED_VFOLDERS
 from ..types import UserScope
 from .base import (
@@ -480,10 +480,6 @@ async def prepare_vfolder_mounts(
     vfolder configurations, and the given user scope.
     """
 
-    # Fast-path for empty requested mounts
-    if not requested_mounts:
-        return []
-
     requested_vfolder_names: dict[str, str] = {}
     requested_vfolder_subpaths: dict[str, str] = {}
     requested_vfolder_dstpaths: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/vfolder.py
```diff
@@ -530,13 +526,24 @@ async def prepare_vfolder_mounts(
         extra_vf_conds=extra_vf_conds,
     )
 
+    # Fast-path for empty requested mounts
+    if not accessible_vfolders:
+        return []
+    from icecream import ic
```
fregataa, line 532: is this used for debugging? `ic` is used on 559 line!

src/ai/backend/manager/models/vfolder.py
```diff
@@ -530,13 +526,24 @@ async def prepare_vfolder_mounts(
...
+
     # for vfolder in accessible_vfolders:
     for key, vfolder_name in requested_vfolder_names.items():
         for vfolder in accessible_vfolders:
             if vfolder['name'] == requested_vfolder_names[key]:
                 break
-        else:
-            raise VFolderNotFound(f"VFolder {vfolder_name} is not found or accessible.")
+            if vfolder['name'] not in requested_vfolder_names:
+                raise VFolderNotFound(f"VFolder {vfolder_name} is not found or accessible.")
```
fregataa, line 546:
```
        if vfolder_name not in [vfolder['name'] for vfolder in accessible_vfolders]:
            raise VFolderNotFound(f"VFolder {vfolder_name} is not found or accessible.")
```

how about this?
rather iterating all `accessible_vfolders`, using list comprehension seems clear

fregataa (commented):
thanks! I leave comments

src/ai/backend/manager/models/vfolder.py
achimnol: @lizable If you want to suggest `icecream` as a tool for development setups, please add it the setup.cfg's `options.extra_require.dev` section but **_remove_** in the final PR code for review & production.

FYI: To debug-print something with its own expression, try `print(f"{expr=}")` in Python. ([ref](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals))
image

lizable: My mistakes. I'll fix it right away.

src/ai/backend/manager/models/vfolder.py
lizable: Yes, It seems like the main purpose of the "original" commit was iterating the `name` field for accessible_vfolders.
I'll apply it.

adrysn (approved):
Tested and it is working.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -530,13 +526,24 @@ async def prepare_vfolder_mounts(
         extra_vf_conds=extra_vf_conds,
     )
 
+    # Fast-path for empty requested mounts
+    if not accessible_vfolders:
+        return []
```
achimnol, line 531: I think we should handle this case as well.
Instead of directly applying my suggestion, could we simplify the matching?
```suggestion
    if not accessible_vfolders:
        if requested_vfolder_names:
            raise VFolderNotFound(f"VFolder {requested_vfolder_names[0]} is not found or accessible.")
        else:
            return []
```

src/ai/backend/manager/models/vfolder.py
```diff
@@ -530,13 +526,24 @@ async def prepare_vfolder_mounts(
...
+
+    # add automount folder list into requested_vfolder_names
+    # and requested_vfolder_subpath
+    for vfolder in accessible_vfolders:
+        if vfolder['name'].startswith('.'):
+            requested_vfolder_names.setdefault(vfolder['name'], vfolder['name'])
+            requested_vfolder_subpaths.setdefault(vfolder['name'], '.')
```
achimnol, line 538: 👍🏼

achimnol, line 538: 👍🏼

src/ai/backend/manager/models/vfolder.py
achimnol: I think Sanghun's suggestion is good. The original inner for-loop is a boilerplate.

achimnol (changes requested):

achimnol (approved):

PR #596, opened 2022-05-24
fix: Add an explicit check for redis connection

rapsealk:
This PR includes a code to check redis connection explicitly.
This will help diagnosis of both permanent failures due to an incorrect server URL and temporary network failures.

Refs
- lablup/backend.ai#424
- lablup/backend.ai-common#139


achimnol:
Replaced with lablup/backend.ai#512.

PR #597, opened 2022-05-31
feat: add a GraphQL query to get the information of a virtual folder by ID

adrysn:


adrysn:
Close this issue since we determined to add this feature to 22.06 or later.

